PR #2, opened 2019-03-29
Ensure librpminspect is linked against libdl

davide125:
Add a missing `libdl` dependency, which fixes a build with `--enable-static --disable-shared`. Specifically, this is needed by `src/librpminspect/inspect_elf.c` as it calls `dlopen()`.


dcantrell:
Very nice, thank you.  Also, sorry this took so long to merge.  github is not sending me notifications in any useful manner with a PR shows up.  This isn't the only project I'm having that happen on.

PR #7, opened 2019-04-23
Fix the tests

dashea:
`make check` was returning success whether the tests actually succeeded or not. Fix the test running code so that failures fail, fix or remove the failing tests, and also make sure that the CUnit output is displayed in Travis on test failure.


dashea:
Couple more commits to skip the tty_width test in travis

PR #10, opened 2019-05-30
Badwords and misc tests

dashea:
Change the badwords implementation to not use regexes, since there's no real reason to use them and it just adds another layer of complexity.

Also clean up a couple of things with the tests.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,7 +16,7 @@ before_install:
 
 script:
   - $DOCKER_RUN ./autogen.sh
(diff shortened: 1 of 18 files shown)


dcantrell:
I like it and I like the improvements for the test suite.  I've never used -no-install, but that's exactly what I needed.

Test suite fails, seems to gripe about CFLAGS= on the $DOCKER_RUN line.  Command not found.  Preface it with env?

PR #31, opened 2019-08-22
doc: fix typos

jwilk:


dcantrell:
Thanks for the patch!  This file was generated with the koji command.  I did this:

koji list-api > koji-api.txt

If you want to fix the typos in koji as well, I think they would appreciate it.

PR #33, opened 2019-08-30
Fix SEGV in get_product_release

cipherboy:
When the user passes an obviously bogus release like:

    rpminspect 42

`rindex(after, '.')` returns `NULL`, which then gets incremented to 0x1 and
assigned to `pos`. `strdup(pos)` then segfaults prior to the `NULL` check on
`after_product`.

I might think about the error message some more but that should be a later PR. I need to understand more what "build" means in the context of rpminspect (e.g., what artifacts it can be given -- it looks like package names and local RPMs) and if there's a better way of informing the user about their mistake that makes it clear what is wrong with their argument and what they could pass instead.

### DIFF
--- a/src/rpminspect/rpminspect.c
+++ b/src/rpminspect/rpminspect.c
@@ -81,9 +81,18 @@ static char *get_product_release(const char *before, const char *after) {
 
     assert(after != NULL);
 
-    pos = rindex(after, '.') + 1;
-    after_product = strdup(pos);
+    pos = rindex(after, '.');
+    if (!pos) {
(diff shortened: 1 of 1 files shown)


dcantrell:
Nice catch.  I do need to move the release determination block to after collecting builds since this won't work for specifying a Koji build ID on the command line.

I would like to add tests for this in an integration test suite.  For me, that's just a bunch of shell scripts that run the program and test different conditions.  I'll start the suite with a simple test and you can see how to fill things in.

PR #34, opened 2019-08-30
Add regression test for segfault

cipherboy:
Ensure that builds without product releases don't cause a segfault. As
the text of the error might change in the future (or, 42 might become a
valid build id), only check that we get a non-SEGV return code and don't
check program output.

Regression test for #33. 


cipherboy:
ACK, updated. :)

PR #35, opened 2019-08-30
Fix config file detection

cipherboy:
When a configuration file is passed on the command line, it should
always be used. In particular, that should mean erring out and telling
the user that their parameter was bad. When no config parameter was
passed, we should use the default if it exists. However, since
rpminspect is a new package (and choice of default `rpminspect-data-*`
depends on the distro), we should inform the user what they're missing.

The issue with the old behavior is, if an explicit configuration file was passed *and* the user lacked `rpminspect-data-fedora`, they'd default to using a file which doesn't exist. :-) 

### DIFF
--- a/src/rpminspect/rpminspect.c
+++ b/src/rpminspect/rpminspect.c
@@ -397,9 +397,22 @@ int main(int argc, char **argv) {
         }
     }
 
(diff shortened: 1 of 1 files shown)


cipherboy:
ACK, updated.

PR #38, opened 2019-08-30
Add IPv6 blacklist check

cipherboy:
Use the new tests:ipv6_blacklist from the config file to enable checking
of IPv4-only function usage. When these functions are used, it likely
means the program isn't compatible with IPv6 addresses, which are
becoming increasingly more widely used.

Depends on https://github.com/rpminspect/rpminspect-data-fedora/pull/1. 

### DIFF
--- a/lib/init.c
+++ b/lib/init.c
@@ -477,6 +477,13 @@ int init_rpminspect(struct rpminspect *ri, const char *cfgfile) {
         parse_list(tmp, &ri->forbidden_directories);
     }
 
+    tmp = iniparser_getstring(cfg, "tests:elf_ipv6_blacklist", NULL);
+    if (tmp == NULL) {
+        ri->ipv6_blacklist = NULL;
+    } else {
+        parse_list(tmp, &ri->ipv6_blacklist);
+    }
+
     /* If any of the regular expressions fail to compile, stop and return failure */
     if (add_regex(cfg, "tests:elf_path_include", &ri->elf_path_include) != 0) {
         return -1;
--- a/lib/inspect_elf.c
+++ b/lib/inspect_elf.c
@@ -699,6 +699,73 @@ cleanup:
     return result;
 }
 
+/* Check for binaries that use blacklisted functions which don't support IPv6.
+ * This could indicate broken support for IPv6. */
+static bool check_ipv6(struct rpminspect *ri, Elf *after_elf, const char *localpath, const char *arch)
+{
+    string_list_t *after_symbols = NULL;
+    string_list_t *used_symbols = NULL;
+    string_list_t *sorted_used = NULL;
+    string_entry_t *iter = NULL;
+
+    char *msg = NULL;
+    bool result = true;
+
+    FILE *output_stream = NULL;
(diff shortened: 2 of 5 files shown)


cipherboy:
@dcantrell Any thoughts about this error? I don't see it on my system. Perhaps it is due to my config changes?

dcantrell:
The Travis-CI failure is due to a declared but unused variable.  Travis-CI uses -Werror when building, so it's a bit more pedantic.

dcantrell:
OK, overall this PR looks good.  But the more I think about it, the more I think this could just be a generic "forbidden functions" check and not strictly for IPv6.  The config file setting could just carry a list of functions that are forbidden.  Thoughts?

cipherboy:
We could add a generic "forbidden functions" check, but my argument is IPv6 should be different than a forbidden functions check. 

Thoughts:
 - IPv4 / IPv6 have mirrored functions and constants. Take FreeRADIUS as an example: they explicitly handle IPv4 addresses with separate functions (that aren't IPv4 aware and don't need to be) than IPv6 addresses. So our usage, while... unique, isn't non-compliant just because we use IPv4-only functions. 
 - IPv6 should be maintainer-overridable, but some forbidden functions might need other processes to review.
 - All usages of IPv6 could be reviewed and grouped together, but other forbidden functions should be reviewed and waived individually (and, policy might dictate re-review each time). 

If we enable the above points for IPv6 in a generic forbidden-functions mechanism, the configuration format might get complex. 

Perhaps I'm thinking about this wrong though. :) 



And, good catch about the unused variable, I missed that in the middle of the output.

dcantrell:
That's all sound and as I think about it more, one of the goals of rpminspect should be to provide more fine-grained and grouped review of builds.  Grouping IPv6 checks together is important (it has been in the past from a business perspective).  Plus, just checking for the existence of a symbol or not is not particularly useful.  We should do more detailed analysis if we're even bothering with checks anyway.

And yes, I do not want the config file to grow way too complicated.

dcantrell:
On the grouping note, I would like to have a distinct 'ipv6' inspection separate from the ELF inspection.  To allow maintainers to enable or disable the ipv6 checks per run.  In inspect.h, just don't use 14.  I've already claimed it.  :)

dcantrell:
Any update on this?  Not sure if it was clear from my last comment, but I like the approach you took and don't mind it in the inspect_elf.c code.  The code has changed around a bit.  Can you get this updated so I can merge it?

cipherboy:
Rebased this on top of master per IRC discussion. Still working on verifying my changes applied cleanly.

cipherboy:
Testing with:

```
$ ./rpminspect freeradius-3.0.19-5.fc31
~snip~
12) /usr/lib/freeradius/rlm_sql.so may use functions unsuitable for IPv6 support on armv7hl 
Result: VERIFY
Waiver Authorization: Anyone

Screendump:
IPv4-only symbols used:
	inet_addr


Suggested Remedy: Please review all usages of IPv4-only functions and ensure IPv6 compliance.
~snip~
```

So I think we're good to go :)

PR #55, opened 2019-10-11
Fix segfault when path to RPM contains ..

cipherboy:
When the path to the before or after RPM contained '..' in it,
rpminspect would previously segfault:

    $ rpminspect ../koji/resteasy-3.0.19-9.fc30.noarch.rpm resteasy-3.0.26-1.fc32.noarch.rpm
    *** Error extracting /var/tmp/rpminspect/local.6nixV7/before/../koji/resteasy-3.0.19-9.fc30.noarch.rpm: Path contains '..'
    Segmentation fault (core dumped)

This is undesirable. To prevent the segfault, fix error handling in
add_peer and get_rpm_info in order to detect when extract_rpm fails.

This has the side effect of showing a bogus second error message though:

    $ rpminspect ../koji/resteasy-3.0.19-9.fc30.noarch.rpm resteasy-3.0.26-1.fc32.noarch.rpm
    *** Error extracting /var/tmp/rpminspect/local.hcZD52/before/../koji/resteasy-3.0.19-9.fc30.noarch.rpm: Path contains '..'
    *** Error gathering build ../koji/resteasy-3.0.19-9.fc30.noarch.rpm: No such file or directory
    *** Failed to gather specified builds.

But at least it doesn't segfault any more.

---

Original issue I began filing before figuring out the (partial) solution:

Leaving this one to the expert... :)

`extract_rpm` errors when the path includes `..`. I think this is partly desirable in case the packaged RPM includes it, but in my case, its when specified on the command line to `rpminspect`:

```
$ ~/GitHub/cipherboy/rpminspect/build/rpminspect ../koji/resteasy-3.0.19-9.fc30.noarch.rpm resteasy-3.0.26-1.fc32.noarch.rpm
*** Error extracting /var/tmp/rpminspect/local.6nixV7/before/../koji/resteasy-3.0.19-9.fc30.noarch.rpm: Path contains '..'
Segmentation fault (core dumped)
```

Clearly we at least shouldn't segfault.

The error message (`Error extracting...`) comes from [`extract_rpm`](https://github.com/rpminspect/rpminspect/blob/master/lib/files.c#L245-L251). However, the segfault happens much later:

```
Program received signal SIGSEGV, Segmentation fault.
0x00007ffff7fb85ae in inspect_removedfiles (ri=0x409600 <ri>) at ../lib/inspect_removedfiles.c:158
(gdb) backtrace
#0  0x00007ffff7fb85ae in inspect_removedfiles (ri=0x409600 <ri>) at ../lib/inspect_removedfiles.c:158
#1  0x000000000040387b in main (argc=3, argv=0x7fffffffd6a8) at ../src/rpminspect.c:572
```

In particular, `inspect_removedfiles` seems to assume `before_files` [is non-NULL](https://github.com/rpminspect/rpminspect/blob/master/lib/inspect_removedfiles.c#L158). Fine, that's an assumption we make in a lot of places.

The issue seems to be that `extract_rpm` has no way of returning an error up to `main`. Starting at the end: 

 - `extract_rpm` is called by [`add_peer`](https://github.com/rpminspect/rpminspect/blob/master/lib/peers.c#L142). This doesn't check that the result is non-NULL but also is a void function so nothing can really be done here anyways. 
 - Even if `add_peer` now returns an error value, [`get_rpm_info`](https://github.com/rpminspect/rpminspect/blob/master/src/builds.c#L103) needs to be updated to check it.
 - The rest seems fine.

-----

I'm guessing `errno` is getting set somewhere and we just need to clear it? 

Signed-off-by: Alexander Scheel <ascheel@redhat.com>

### DIFF
--- a/lib/peers.c
+++ b/lib/peers.c
@@ -68,7 +68,7 @@ void free_rpmpeer(rpmpeer_t *peers) {
 /*
  * Add the specified package as a peer in the list of packages.
  */
-void add_peer(rpmpeer_t **peers, int whichbuild, bool fetch_only, const char *pkg, Header *hdr) {
+int add_peer(rpmpeer_t **peers, int whichbuild, bool fetch_only, const char *pkg, Header *hdr) {
     rpmpeer_entry_t *peer = NULL;
     bool found = false;
     const char *newname = NULL;
@@ -97,6 +97,15 @@ void add_peer(rpmpeer_t **peers, int whichbuild, bool fetch_only, const char *pk
         existingarch = NULL;
         existingsrc = false;
 
+        if (existingname != NULL && !strcmp(pkg, existingname)) {
+            return 0;
+        }
+    }
+
+    /* Second, if we don't have this peer, try to add it */
+    existingname = NULL;
+
+    TAILQ_FOREACH(peer, *peers, items) {
         if (whichbuild == BEFORE_BUILD && peer->after_rpm != NULL) {
             existingname = headerGetString(peer->after_hdr, RPMTAG_NAME);
             existingarch = headerGetString(peer->after_hdr, RPMTAG_ARCH);
@@ -124,7 +133,7 @@ void add_peer(rpmpeer_t **peers, int whichbuild, bool fetch_only, const char *pk
         if ((peer = calloc(1, sizeof(*peer))) == NULL) {
             fprintf(stderr, "*** failed to allocate new peer peer\n");
             fflush(stderr);
-            return;
+            return 1;
         }
     }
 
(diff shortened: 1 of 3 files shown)


cipherboy:
Guess my assumption that `extract_rpm(...) == NULL <-> erred` isn't correct. I guess we could make `extract_rpm` return an error now and move the list to an argument... thoughts?

dcantrell:
Test suite is working again but you need to update this PR against master before it can run again.

dcantrell:
Tests are failing right now, but it's my fault.  Pay no attention to them.  Patch set looks good.

PR #100, opened 2020-02-12
updating imports to use libcap, matching BuildRequires in specfile

### LINKED ISSUE #99: COPR builds are failing due to missing cap-ng.h
The COPR builds have been failing since February 1:
https://copr.fedorainfracloud.org/coprs/dcantrel/rpminspect/builds/

They're failing in compilation due to a missing header file:
```
../lib/files.c:30:10: fatal error: cap-ng.h: No such file or directory
   30 | #include <cap-ng.h>
      |          ^~~~~~~~~~
```
I assume this is due to the intended change from cap-ng to libcap that changed in the specfile but may not have changed everywhere in the code.

tflink:
I think this fixes #99 by updating the imports in file.c and inspect_ownership.c
The change builds locally and in mock. Some quick tests locally work with this patch.

### DIFF
--- a/lib/files.c
+++ b/lib/files.c
@@ -27,7 +27,7 @@
 #include <unistd.h>
 #include <libgen.h>
 
-#include <cap-ng.h>
+#include <sys/capability.h>
 
 #include <rpm/header.h>
 #include <rpm/rpmtd.h>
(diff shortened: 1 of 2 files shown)


dcantrell:
Yep, this is correct.  Should have changed those over when I moved over to libcap.  Thanks!

PR #120, opened 2020-03-30
Add tests for shellsyntax for /bin/csh

sturivny:


dcantrell:
csh is not a package.  It's provided as a symlink in the tcsh package.

PR #121, opened 2020-03-30
Add tests for shellsyntax for /bin/tcsh

sturivny:


dcantrell:
This one has merge conflicts for Dockerfile.test.

Also it looks like the tests failed due to tcsh syntax.

sturivny:
Fixed merge conflict.
What do you mean by `tests failed due to tcsh syntax`?

dcantrell:
The invalid_tcsh example script you provided is passing "tcsh -n".  I'm not sure why.  There may be an example script in the tcsh source code that is invalid.

dcantrell:
Here's an example:

```
#!/bin/tcsh
if [ $name != $user ) then
    echo "Your name isn't your username"
else
    echo "Your name is your username"
endif
```

If I save that to test.tcsh and run `tcsh -n test.tcsh`, I get:

```
$ tcsh -n test.tcsh 
Too many )'s.
```

And the exit code is non-zero.  The same type check can be used for csh.

sturivny:
So, I'll update PRs with a script that not passed `tcsh -n`.
Thanks for the explanation.

sturivny:
Done

PR #162, opened 2020-07-09
ci: Add ShellCheck to lint shell scripts

jeremycline:
ShellCheck[0] is a great linting tool for shell scripts and catches a
number of common problems.

[0] https://www.shellcheck.net/

### DIFF
--- a/.github/workflows/ci-shell-style.yml
+++ b/.github/workflows/ci-shell-style.yml
@@ -1,0 +1,17 @@
+name: Shell Style Checks
+
+on:
+    push:
+        branches: [master]
+    pull_request:
+        branches: [master]
+
+jobs:
+    shellcheck:
+        runs-on: ubuntu-latest
+        # Ubuntu's ShellCheck is old enough to not have --severity
+        container: fedora:latest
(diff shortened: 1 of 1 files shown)


jeremycline:
> This looks good, with two comments:
> 
>     * Can you make the name something like "Shell Script Style"

I was planning to stash all the style-related checks in there, but I'm happy to have them each in individual files if you'd prefer that. It's just a bit more boilerplate for each check.

> 
>     * Please use 4 space indentation in yaml files.  I can't read it otherwise.  I'm old.
> 

Happy to do so

dcantrell:
> > This looks good, with two comments:
> > ```
> > * Can you make the name something like "Shell Script Style"
> > ```
> 
> I was planning to stash all the style-related checks in there, but I'm happy to have them each in individual files if you'd prefer that. It's just a bit more boilerplate for each check.

I'd prefer to break each one in to a separate action.

> > ```
> > * Please use 4 space indentation in yaml files.  I can't read it otherwise.  I'm old.
> > ```
> 
> Happy to do so

Thanks.

PR #164, opened 2020-07-10
Style the Python code with Black

jeremycline:
This is a large commit, which is unfortunate for review, but its advantage is only one commit needs to be added to the revision ignore list for blame.

### DIFF
--- a/.git-blame-ignore-revs
+++ b/.git-blame-ignore-revs
@@ -1,0 +1,2 @@
+# Migrate Python code to Black
+4a192214ab27722af0f63cd86328ea25665c5146
--- a/contrib/report-json2html.py
+++ b/contrib/report-json2html.py
@@ -75,12 +75,12 @@ def load_json_report(path):
     return data
 
 
-if __name__ == '__main__':
+if __name__ == "__main__":
     if len(sys.argv) < 2:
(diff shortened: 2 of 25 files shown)


.git-blame-ignore-revs
```diff
@@ -0,0 +1,2 @@
+# Migrate Python code to Black
```
dcantrell, line 1: What does this file do?

jeremycline: I left a note in the commit that formats the code on usage, you can pass it to git-blame so it ignores that particular commit. This way, if you're blame'ing around and it points at the reformat commit, you can look at that, get instructions on how to ignore it, and continue looking at blame info as if the commit never existed. Git documentation is https://git-scm.com/docs/git-blame#Documentation/git-blame.txt---ignore-revs-fileltfilegt

dcantrell: OK, got it.  Nice.  Never used that functionality on git blame.

jeremycline:
Yeah, I'm definitely not a fan of some of its style choices, but as much as I dislike them I dislike correcting people's style even more. Telling contributors to just add a pre-commit hook or similar is worth it.

PR #167, opened 2020-07-13
Upload coverage report to codecov

jeremycline:
codecov provides a nice online code coverage tracker (with a badge!) so
that code coverage can be tracked over time and examined from the
browser. All that's necessary to enable this add the upload token[0] to
the repository secrets (Settings -> Secrets) with the name CODECOV_TOKEN

[0] https://codecov.io/gh/rpminspect/rpminspect/

An example report: https://codecov.io/gh/jeremycline/rpminspect/tree/c4866f0848cbf5516b17c6e9415e56ec9cbb503e

### DIFF
--- a/.github/.codecov.yml
+++ b/.github/.codecov.yml
@@ -1,0 +1,3 @@
+ignore:
+  - "test/"
+  - "build/"
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -42,7 +42,9 @@ jobs:
             # Generate code coverage reports (requires
             # -Db_coverage=true on 'meson setup')
             - name: coverage
-              run: ninja -C build coverage
+              run: |
+                  ninja -C build coverage
+                  curl -s https://codecov.io/bash | bash
 
     ubuntu:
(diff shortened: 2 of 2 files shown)


dcantrell:
Ahh, very cool.  So I was trying to get it to upload to coveralls.something because that's what was previously maybe happening.  I don't know much about these code coverage reporting sites, but I wanted it working.  Coveralls appeared to take a dump all over whatever I tried to do so I gave up hoping that someone else would fix it.  And that happened!  Thanks!

jeremycline:
I've mostly used codecov although I think the feature sets are about the same. It's worth noting I've had codecov produce some... odd results on occasion (typically due to a combination of merging coverage reports combined with slow/failing CI), but it mostly works. I also just noticed that they've got a GitHub Action (https://github.com/marketplace/actions/codecov) so if the bash uploader starts causing problems that might be something to look at.

Finally, it's possible to have it comment with the coverage report of PRs and add coverage requirements to the list of PR checks, if you're so inclined. IIRC you need to install https://github.com/marketplace/codecov so it has permissions to comment and so on.

PR #168, opened 2020-07-14
Tidy up the test base classes

jeremycline:
While skimming through the test interface to figure out what's required to implement a test, there were a few oddities that made me jump around the code a bit more than is strictly necessary. These changes are strictly equivalent, but somewhat more (in my opinion) idiomatic and clear.


dcantrell:
Very nice, thank you.  I see you are wading through my "traceback driven development" coding style.  Thanks for fixing this stuff up.

PR #169, opened 2020-07-14
Improve the error reporting for test result checking

jeremycline:
While developing a new test, I provided the wrong waiver_auth and got a
test that failed with:

Traceback (most recent call last):
  File "/home/jcline/devel/rpminspect/test/baseclass.py", line 419, in
  runTest
      self.assertTrue(
      AssertionError: False is not true

It is indeed true that False is not true, but I had to do a bit of
digging to determine what was wrong with my test.

This changes check_results to raise an AssertionError with details about
what exactly was wrong with the results. For example, the issue that generated
the above error now generates:

Traceback (most recent call last):
  File "/home/jcline/devel/rpminspect/test/baseclass.py", line 414, in runTest
    check_results(self.results, self.label, self.result, self.waiver_auth)
  File "/home/jcline/devel/rpminspect/test/baseclass.py", line 67, in check_results
    raise AssertionError(
AssertionError: Expected result=VERIFY with waiver authorization=Not Waivable in [
    {
        "message": "/some/file grew by +2000% on x86_64",
        "result": "VERIFY",
        "waiver authorization": "Anyone"
    }
]

### DIFF
--- a/test/baseclass.py
+++ b/test/baseclass.py
@@ -53,21 +53,21 @@ class MissingRpminspectConf(Exception):
 # have at least one result that matches the result and waiver_auth we
 # were expecting.
 def check_results(results, label, result, waiver_auth):
-    ret = False
-
     if results == {} or results == [] or results is None:
-        return False
+        raise AssertionError("JSON test result is empty")
 
     if label not in results:
-        return False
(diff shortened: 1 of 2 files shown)


dcantrell:
If it was hard to write, it should be hard to read!  j/k, thanks

PR #170, opened 2020-07-14
Add basic filesize inspection tests and fix a bug they revealed

jeremycline:
Feel free to squash these if you'd prefer all commits to have passing tests, they're broken up mainly so it's easy to git-revert and verify the tests begin passing with the change.

### DIFF
--- a/test/test_filesize.py
+++ b/test/test_filesize.py
@@ -1,0 +1,80 @@
+#
+# Copyright (C) 2020  Red Hat, Inc.
+# Author(s):  Jeremy Cline <jcline@redhat.com>
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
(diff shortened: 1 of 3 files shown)


test/test_filesize.py
```diff
@@ -0,0 +1,83 @@
...
+
+        # Tests have threshold configuration of 20%
+        self.before_rpm.add_installed_file(
+            "/some/file", rpmfluff.SourceFile("file", "a" * 5)
+        )
+        self.after_rpm.add_installed_file(
+            "/some/file", rpmfluff.SourceFile("file", "a" * 6)
+        )
+
+        self.exitcode = 1
```
dcantrell, line 38: You shouldn't need to set self.exitcode in these tests.  That should be handled by baseclass.py

jeremycline: Ah yes I see. I couldn't think of any good puns so I leave any good fake filenames up to you

PR #199, opened 2020-09-29
'patches' inspection, filecmp(), and test cases for 'changedfiles' and 'patches'

dcantrell:


include/inspect.h
```diff
@@ -566,4 +577,7 @@ bool inspect_doc(struct rpminspect *ri);
 
 #define DESC_DOC _("Report changes in %doc marked files.  Files that have been added or removed as documentation files, for instance.")
 
+#define DESC_PATCHES _("Inspects all patches defined in the spec file and reports changes between builds.  At the INFO level, rpminspect reports a diffstat(1) and patch size changes.  If thresholds are reached regarding a change in the patch size or the number of files the patch touches, rpminspect reports the change at the VERIFY level unless the comparison is for a rebase.  The configuration file can also list patch names that rpminspect should ignore during the inspection.")
```
jimbair, line 580: Small nitpick:

"At the INFO level, rpminspect reports a diffstat(1) and patch size changes."

I would change that to:

"At the INFO level, rpminspect reports diffstat(1) and patch size changes."

dcantrell: That's reasonable.  I'll add it in the next PR.  Want to get this one merged.

jimbair: Agreed.

jimbair:
Looks good to me; I'm no C expert, but everything reads well and makes sense to me so long as it's not deep in the weeds. :) I say merge away.

PR #206, opened 2020-10-01
Implement the 'virus' inspection and add test cases for it.

dcantrell:
The virus inspection checks every file in the after build (or the only
build if you are running rpminspect that way) using libclamav and
reports if a virus was found.  This inspection is expensive, but
that's how clamav works.  The inspection can be disabled via profiles,
the command line, or local configuration as needed.

If a virus is found, rpminspect reports the file that has the
suspected virus as well as the virus name returned by libclamav.  The
result severity level is BAD.


jimbair:
Looks good to me; nothing to nitpick here. :) +1

PR #266, opened 2021-01-15
[ci] Build and install 'rc' from source on opensuse-leap

dcantrell:
opensuse-leap lacks an installable package for the 'rc' shell.  What I
was doing was fetching it from Fedora's RPM repos, then using rpm2cpio
to unpack the package manually and just copy the 'rc' executable in
place.  This is failing now because of generally yum unpleasantries.
Since rc is small, just download the source from Debian and build it
within the instance.  I'm already doing that for mandoc, so I can do
that for rc.

### DIFF
--- a/osdeps/opensuse-leap/post.sh
+++ b/osdeps/opensuse-leap/post.sh
@@ -29,28 +29,18 @@ make lib-install
 cd ${CWD}
 rm -rf mandoc.tar.gz ${SUBDIR}
 
-# Download 'rc' from Fedora because OpenSUSE does not provide it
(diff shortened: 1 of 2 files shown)


jimbair:
Is it worth adding a check if the URL fetch works cleanly? I assume if cURL is happy, then make is probably safe. Or maybe just check make?

Otherwise, I think this method works good considering the limitations.

dcantrell:
I'd like to let it run this way since if it fails, the CI job would fail anyway.  If it's a regular problem that requires some retries or something, then I suppose we could make this block smarter.

Issue #303, opened 2021-02-24
rpminspect stuck in the middle of the testing


ukulekek:
Hello, we have recently received an update of rpminspect on staging and it's broken. I believe the issue reproduces with any package, I tried 4 different packages.  

The issue has 2 different behaviors:  

With `-v` option it stuck in the middle of testing and never finishes:
```
$ rpminspect-redhat -v openstack-ironic-16.2.1-0.20210223151242.999e340.el8
Downloading <truncated>
Running license inspection...        FAIL
Running emptyrpm inspection...       pass
Running metadata inspection...       pass
Running manpage inspection...        pass
Running xml inspection...            pass
corrupted size vs. prev_size while consolidating
```
I believe it's happening in the elf inspection step because the previous version of rpminspect reports elf inspection just after that:
```
$ rpm -q rpminspect
rpminspect-1.3-0.1.202101272033git.fc32.x86_64
$ rpminspect-redhat -v openstack-ironic-16.2.1-0.20210223151242.999e340.el8
Downloading <truncated>
Running license inspection...        FAIL
Running emptyrpm inspection...       pass
Running metadata inspection...       pass
Running manpage inspection...        pass
Running xml inspection...            pass
Running elf inspection...            pass
Running desktop inspection...        pass
Running disttag inspection...        pass
Running specname inspection...       pass
Running modularity inspection...     pass
Running javabytecode inspection...   pass
Running ownership inspection...      pass
<truncated>
$ echo $?
0
```

Without `-v` option it reports Segmentation fault:
```
$ rpminspect-redhat openstack-ironic-16.2.1-0.20210223151242.999e340.el8
/usr/bin/rpminspect-redhat: line 2:    39 Segmentation fault      (core dumped) /usr/bin/rpminspect -c /usr/share/rpminspect/redhat.yaml $*
$ echo $?
139
```

* rpminspect-1.3-0.1.202102231714git.fc32.x86_64
* librpminspect-1.3-0.1.202102231714git.fc32.x86_64
* rpminspect-data-redhat-1.4-0.1.202102232053git.fc32.noarch
* Fedora-32

dcantrell:
Looking at this one now.

dcantrell:
Definitely a bug.  I found it in two lines in init_elf_data() in readelf.c.  I had duplicated a cleanup block which meant the function was calling elf_end() twice and libelf was not returning.  Fix coming momentarily.

PR #304, opened 2021-02-24
[fix] Remove duplicate elf_end() call in init_elf_data() (#303)

dcantrell:
elf_end() and close() were possibly called twice for successful runs
of init_elf_data().  Calling elf_end() more than once on an opened
libelf entity is undefined.  It could crash or it could run forever.
Point is, don't do it.


jimbair:
Looks good to me; drop them dupes

Issue #308, opened 2021-03-01
rpminspect: archive_read_next_header(): No such file or directory


morucci:
Hi David,

I saw that rpminspect crashed during grub2 packages inspection with

```
2021-02-22 18:59:58.416295 | container | Running specname inspection...       pass
2021-02-22 18:59:58.416327 | container | Running modularity inspection...     pass
2021-02-22 18:59:58.417186 | container | Running javabytecode inspection...   pass
2021-02-22 19:00:01.775670 | container | rpminspect: archive_read_next_header(): No such file or directory
2021-02-22 19:00:01.775741 | container | rpminspect: ../lib/inspect_changedfiles.c:228: changedfiles_driver: Assertion `before_uncompressed_file != NULL' failed.
2021-02-22 19:00:01.780042 | container | /usr/bin/rpminspect-fedora: line 3:  2407 Aborted                 /usr/bin/rpminspect -c /usr/share/rpminspect/fedora.yaml $*
2021-02-22 19:00:02.033798 | container | ERROR
2021-02-22 19:00:02.034303 | container | {
2021-02-22 19:00:02.034509 | container |   "delta": "0:00:07.071133",
2021-02-22 19:00:02.034589 | container |   "end": "2021-02-22 19:00:01.781142",
2021-02-22 19:00:02.034654 | container |   "msg": "non-zero return code",
2021-02-22 19:00:02.034716 | container |   "rc": 134,
2021-02-22 19:00:02.034842 | container |   "start": "2021-02-22 18:59:54.710009"
2021-02-22 19:00:02.034906 | container | }
```

See for the full log: https://fedora.softwarefactory-project.io/zuul/build/7440ae99179a4abf9bb057bc4b971383/log/job-output.txt#1190

Any idea about this issue ?

Thanks in advance.

dcantrell:
I am unable to reproduce this problem locally.  A couple of things:

- What version of rpminspect are you using?  I have 1.3.1 built and available in updates-testing, but at least version 1.2 should be used.
- What builds are you comparing here?  Koji told me no such build for grub2-2.04-35.fc35.

Just running the latest stable rpminspect on grub2-2.04-35.fc34 ran fine.  Please advise.

morucci:
I'm able to reproduce out of the CI, locally.

grub2-2.04-35.fc35 was built by the CI from that PR https://src.fedoraproject.org/rpms/grub2/pull-request/13

Basically to reproduce:
```
[...]

PR #312, opened 2021-03-02
[fix] Handle compressed but otherwise empty man pages (#308)

dcantrell:
A couple of fixes in this patch.  First, make sure uncompress_file()
handles compressed but empty files.  Because libarchive returns
ARCHIVE_EOF on archive_read_next_header() when the payload is empty,
my function was erroring out and returning NULL.  That caused an
assert() failure higher up in the manpage inspection.  So, teach
uncompress_file() how to handle compressed zero-length files.  Since
uncompress_file() creates a temporary file to write the uncompressed
payload to, use fdopen() and fclose() to create a zero-length file in
the case of a compressed empty file.  ARCHIVE_FAILED and ARCHIVE_FATAL
are still handled, but I also treat ARCHIVE_RETRY the same way and
just fail out.  I have not seen ARCHIVE_RETRY triggered in the wild
and in our cases it really only means you can restart everything.  I
thinkg ARCHIVE_RETRY would be useful is streaming from a device or
network but then you'd need some poll() and select() around handling
that case and I kind of want to not do that right now.  Something
tells me I will be back here adding that in the future.

Next is expanding the manpage inspection so it reports empty manpages.
It turns out that libmandoc will successfully validate nothing as
valid mandoc.  So I added a check to see if the uncompressed manpage
is size 0 and if so report that it's probably empty with a VERIFY
warning.  This could be INFO but I image we really don't want packages
including empty man pages so VERIFY feels appropriate here.

### DIFF
--- a/lib/inspect_manpage.c
+++ b/lib/inspect_manpage.c
@@ -31,6 +31,7 @@
 #include <sys/stat.h>
 #include <sys/types.h>
 #include <unistd.h>
+#include <err.h>
 
 #include <rpm/header.h>
 #include <rpm/rpmtag.h>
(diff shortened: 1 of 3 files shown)


jimbair:
Looks good!

PR #377, opened 2021-04-26
[fix] Call mparse_reset() before mparse_readfd()

dcantrell:
A few packages were causing rpminspect to segfault in the manpage
inspection.  I was not calling mparse_reset() before mparse_readfd(),
which was the correct thing to do.  Some things were uninitialized as
a result and if needed during the read and parse could have led to a
segfault.  It did.  Call mparse_reset() to fix things up before
reading and parsing.


jimbair:
Looks good to me

PR #381, opened 2021-04-27
Require/Recommend /usr/bin/annocheck

hroncok:
That is what rpminspect needs, no need to know the package name.

Related: https://bugzilla.redhat.com/show_bug.cgi?id=1941131

### DIFF
--- a/rpminspect.spec.in
+++ b/rpminspect.spec.in
@@ -89,9 +89,9 @@ Requires:       bash
 # The annocheck program is used by the annocheck inspection.  If it is
 # not present, you can disable the annocheck inspection.
 %if 0%{?rhel} >= 8 || 0%{?epel} >= 8 || 0%{?fedora}
-Recommends:     annobin-annocheck
+Recommends:     /usr/bin/annocheck
 %else
-Requires:       annobin-annocheck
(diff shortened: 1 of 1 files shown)


dcantrell:
This looks good.  For future PRs, I would prefer using the 'git commit -s' option to add the Signed-off-by line.

hroncok:
> For future PRs, I would prefer using the 'git commit -s' option to add the Signed-off-by line.

I've used the GitHub edit feature, so I would need to add that sign off line manually.

Out of curiosity, why would you prefer that? What is here to sign off?

dcantrell:
DCO workflow, that's all.  I'm trying to be more consistent across projects.

hroncok:
I haven't even noticed DCO mentioned here anywhere, hence the sign off seems like moot to me. Anyway, thanks for accepting the PR without that line.

dcantrell:
I don't think I have it mentioned anywhere, actually.  It's not a problem with me if you don't send patches with the signed-off-by piece.  I suppose I could add a DCO file in to the project.  I have found some people are generally apathetic towards the DCO and some are very much against it.  I don't really understand the latter, but I don't really mind either way.  If you want to or don't mind doing the DCO thing, awesome.

Issue #382, opened 2021-04-28
segmentation fault with qt5-qtbase package


ukulekek:
Hi, I'm getting segfault with qt5-qtbase package in analysis and assertion failed in comparison mode
```
$ rpminspect-redhat -v  qt5-qtbase-5.15.2-3.el8
Downloading ...
Running license inspection...        pass
Running emptyrpm inspection...       pass
Running metadata inspection...       pass
Running manpage inspection...        pass
/usr/bin/rpminspect-redhat: line 2: 243343 Segmentation fault      (core dumped) /usr/bin/rpminspect -c /usr/share/rpminspect/redhat.yaml $*

$ rpminspect-redhat -v qt5-qtbase-5.15.2-2.el8 qt5-qtbase-5.15.2-3.el8
Downloading ...
Running elf inspection...            pass
Running desktop inspection...        pass
rpminspect: ../lib/macros.c:68: get_specfile_macros: Assertion `spec != NULL' failed.
/usr/bin/rpminspect-redhat: line 2: 289609 Aborted                 (core dumped) /usr/bin/rpminspect -c /usr/share/rpminspect/redhat.yaml $*

```
* rpminspect-1.5-0.1.202104272014git.fc32.x86_64
* librpminspect-1.5-0.1.202104272014git.fc32.x86_64
* rpminspect-data-redhat-1.4-0.1.202104121907git.fc32.noarch

dcantrell:
Oh that's ugly.  I just reproduced this one locally.  I'm on it.

dcantrell:
This is what I get:

```
#0  __strlen_avx2 () at ../sysdeps/x86_64/multiarch/strlen-avx2.S:65
#1  0x00007f0e7a2c1783 in __GI___strdup (s=0x0) at strdup.c:41
#2  0x00007f0e7a4e8d0e in is_xml_well_formed (errors=0x7ffd49e3b518, 
    path=0x19aae30 "/data/bigvar/tmp/rpminspect/local.FIcAPI/after/ppc64le/qt5-qtbase-tests-5.15.2-3.el8.ppc64le//usr/lib64/qt5/tests/auto/corelib/serialization/qxmlstream/XML-Test-Suite/xmlconf/eduni/errata-2e/xmlconf.x"...) at ../lib/inspect_xml.c:83
#3  xml_driver (file=0x19aacb0, ri=0x789ef0) at ../lib/inspect_xml.c:200
[...]

PR #385, opened 2021-04-29
[fix] Handle corrupt compressed files in 'changedfiles' (#382)

dcantrell:
This one took a while to show up, but the reproducer is qtbase (or a
subpackage of it) that includes test suite files.  Among the files is
a corrupted gzipped file.  In changedfiles, I try to uncompress
compressed files to compare the uncompressed content.  This is done to
handle changes in the compression level from build to build.  I was
failing to account for deliberately corrupted compressed files.  This
patch adjusts changedfiles so if a compressed file is corrupted, it
falls back to byte by byte comparison.


jimbair:
Huh; an intentionally corrupted file. That's a good one to check for; nice catch.

PR #396, opened 2021-05-05
Require up-to-date annocheck for RHEL 9.

adelton:


dcantrell:
This updates the requirement for rpminspect running on RHEL-9.  Do we not also want to require version 9.65 on annobin-annocheck on other systems?  With the runners being on Fedora, I would think we want this for Fedora as well.

adelton:
> This updates the requirement for rpminspect running on RHEL-9. Do we not also want to require version 9.65 on annobin-annocheck on other systems? With the runners being on Fedora, I would think we want this for Fedora as well.

I would assume that on Fedora there should be a per-Fedora-version value indicating the minimal annobin version that was used on that particular Fedora.

dcantrell:
> > This updates the requirement for rpminspect running on RHEL-9. Do we not also want to require version 9.65 on annobin-annocheck on other systems? With the runners being on Fedora, I would think we want this for Fedora as well.
> 
> I would assume that on Fedora there should be a per-Fedora-version value indicating the minimal annobin version that was used on that particular Fedora.

We might be talking past each other here.  Since the runners used for RHEL rpminspect are on Fedora, we need to make sure Fedora installs of rpminspect get version 9.65 of annobin-annocheck.  In that case, we would need this block in the spec file:

```
%if 0%{fedora}
Recommends: /usr/bin/annocheck >= 9.65
%endif
```

The patch also changed the Recommends to Requires.  I would prefer to leave this as Recommends since rpminspect does not require annocheck to be present to run, but if you have it the inspection then runs.  Of course, this only applies to systems with versions of RPM that support Recommends.  That's why I have the %if block with the %else clause to just Requires annocheck.

You also changed the dependency from /usr/bin/annocheck to annobin-annocheck.  I had it listed as the executable because the package name varies from EPEL-7 forward, so I found it easier to just require the executable.

With all that, I think what needs to happen in this patch is:

1. A conditional block that adds a Recommends for /usr/bin/annocheck >= 9.65 for RHEL and EPEL >= 9.
2. A conditional block that adds a Recommends for /usr/bin/annocheck >= 9.65 for RHEL and EPEL >= 8. (assuming 9.65 is available for RHEL-8 and EPEL-8, otherwise remove the version test or put the latest release there).
3. A conditional block that adds a Requires for /usr/bin/annocheck for RHEL and EPEL >= 7 since Recommends does not work in packages on that system.
4. A conditional block that adds a Recommends for /usr/bin/annocheck >= 9.65 for Fedora.

I think that should take care of all of the runtime use cases of rpminspect across the Fedora and RHEL releases we need to be concerned with for RHEL-9 build checking.  Can you update the PR?

adelton:
> We might be talking past each other here. Since the runners used for RHEL rpminspect are on Fedora, we need to make sure Fedora installs of rpminspect get version 9.65 of annobin-annocheck.

What are our options for migrating to RHEL runners for the RHEL rpminspect operations, rather than Fedora? It seems like we go into tremendous complexity with this setup which introduces risks to our ability of guaranteeing the required compability of annobin and the matching annocheck.

> The patch also changed the Recommends to Requires. I would prefer to leave this as Recommends since rpminspect does not require annocheck to be present to run, but if you have it the inspection then runs.

I'd like us to gradually become more strict about the tests that we consider part of the standard rpminspect operation on individual platforms. Running annocheck is part of standardised compiler flags testing which should not be allowed to be skipped with weak rpm dependencies.

> You also changed the dependency from /usr/bin/annocheck to annobin-annocheck. I had it listed as the executable because the package name varies from EPEL-7 forward, so I found it easier to just require the executable.

I'm not sure I follow -- there is no `/usr/bin/annocheck` in the original content as shown in https://github.com/rpminspect/rpminspect/pull/396/files.

dcantrell:
> > We might be talking past each other here. Since the runners used for RHEL rpminspect are on Fedora, we need to make sure Fedora installs of rpminspect get version 9.65 of annobin-annocheck.
> 
> What are our options for migrating to RHEL runners for the RHEL rpminspect operations, rather than Fedora? It seems like we go into tremendous complexity with this setup which introduces risks to our ability of guaranteeing the required compability of annobin and the matching annocheck.

That would involve a larger discussion with OSCI.  Can you reach out to osci-list@redhat.com and we can discuss this further.

The main issue is that for a new major release of RHEL, we don't yet have that product to work on.  So having the runners use Fedora is much easier from a production standpoint because we have all of the tools available to us.  It's a better option than using the previous major release of RHEL since all of the tools we want for these jobs may not be available.

I think that moving the runners to RHEL-9 is something we can do after the .0 release.  But we should have a discussion about that with the larger OSCI group.

> > The patch also changed the Recommends to Requires. I would prefer to leave this as Recommends since rpminspect does not require annocheck to be present to run, but if you have it the inspection then runs.
> 
> I'd like us to gradually become more strict about the tests that we consider part of the standard rpminspect operation on individual platforms. Running annocheck is part of standardised compiler flags testing which should not be allowed to be skipped with weak rpm dependencies.

I agree.  In this case, carrying a Requires on annocheck should be added to the rpminspect-data-redhat package to enforce the requirement there.  The rpminspect package itself doesn't require annocheck for the other functionality, which is why it sits at Recommends in that spec file.  But the product data package for rpminspect can enforce more strict requirements.

> > You also changed the dependency from /usr/bin/annocheck to annobin-annocheck. I had it listed as the executable because the package name varies from EPEL-7 forward, so I found it easier to just require the executable.
> 
> I'm not sure I follow -- there is no `/usr/bin/annocheck` in the original content as shown in https://github.com/rpminspect/rpminspect/pull/396/files.

Hmmm, I'm referring to what the spec file says:
https://github.com/rpminspect/rpminspect/pull/396/files

I recently changed that back over to /usr/bin/annocheck to account for the variance in the package name across all product releases.

dcantrell:
I have made this change in the internal rpminspect-data-redhat package so that any systems we use to check RHEL builds using that data package will require annocheck >= 9.65-5

Issue #395, opened 2021-05-05
desktop: ignore (or demote as INFO) missing Exec when TryExec is specified


ptoscano:
The desktop-entry-spec [1] describes a key called `TryExec`. Quoting the spec:

> Path to an executable file on disk used to determine if the program is actually installed. If the path is not an absolute path, the file is looked up in the $PATH environment variable. If the file is not present or if it is not executable, the entry may be ignored (not be used in menus, for example).

This is used, for example, to ship desktop files that enhance a certain executable without a strict dependency on it to avoid extra splitting of packages.

Right now, the `desktop` check reports as VERIFY when the executable specified in `Exec` is not part of any subpackage of the same source. This was reported as #373.

Considering the behaviour hinted by the spec, IMHO another approach could be to ignore the missing `Exec`, or demote its reporting as INFO, in case `TryExec` is specified: XDG menu implementations will check for the existence of the executable in `TryExec`, and if that does not exist then the desktop file will be ignored.

[1] https://specifications.freedesktop.org/desktop-entry-spec/desktop-entry-spec-latest.html

dcantrell:
I like this idea.  Thanks for the information here.  I will extend the desktop inspection to handle the TryExec value in the manner you described.

ptoscano:
Ah, I was actually working on a patch for this.

dcantrell:
Oh, even better.  I will hold off on this and wait for your PR.

PR #401, opened 2021-05-06
[fix] desktop: demote as INFO a missing Exec w/ TryExec (#395)

ptoscano:
According to the desktop-entry-spec [1], menu implementations may ignore a desktop file in case it has a `TryExec` key, and what is mentioned there either does not exist, or it is not executable.

Hence, in case `Exec` does not exist in the inspected RPMs, check whether `TryExec` was specified: if so, then only report the missing `Exec` as INFO, rather than VERIFY.

Includes some mostly no-op code refactoring that makes the fixes much easier to develop & review.

[1] https://specifications.freedesktop.org/desktop-entry-spec/desktop-entry-spec-latest.html

Fixes: #395

### DIFF
--- a/lib/inspect_desktop.c
+++ b/lib/inspect_desktop.c
@@ -169,6 +169,7 @@ static bool validate_desktop_contents(struct rpminspect *ri, const rpmfile_entry
     struct result_params params;
     char *key_exec = NULL;
     char *key_icon = NULL;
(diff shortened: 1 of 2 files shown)


dcantrell:
This is great.  Thank you for the PR on this, I really appreciate it.

PR #464, opened 2021-07-02
fix misleading output from desktop icon check

caolanm:
on libreoffice-startcenter.desktop rpminspect claims it references
icon 'libreoffice' but it really references 'libreoffice-startcenter'
because tmp points to the result of 'Exec' not 'Icon'


dcantrell:
Very nice, thank you.

PR #483, opened 2021-08-05
[ci] Simplify ci.yml with the i686 and x86_64 fedora builds

dcantrell:
Just perform them back to back in the same containers.


dcantrell:
Nope, this is all a bust.

PR #501, opened 2021-08-19
[ci] Add openEuler Linux 20.03 to the extra-ci collection

manas-rust:


manas-rust:
Please review

dcantrell:
I have to disable this because it doesn't work.  openEuler lacks the following packages:

- dash
- kernel-core
- libabigail
- libmandoc-devel
- patchelf
- python3-timeout-decorator
- rc

python3-timeout-decorator can be installed by pip.  The sssd-client package is there, but segfaults on installation.  kernel-core may be CentOS or RHEL specific.  The other packages are available through EPEL, but I cannot enable EPEL on openEuler because no package provides "redhat-release", which epel-release requires.

Without this additional software, openEuler won't work as a CI target.

PR #549, opened 2021-11-09
run desktop-file-validate with --no-hints

caolanm:
Only display unambiguous warnings, not additional helpful tips "about things that might be improved"

f.e.
libreoffice-math.desktop: hint: value "Office;Education;Science;Math;" for key "Categories" in group "Desktop Entry" contains more than one main category; application might appear more than once in the application menu


dcantrell:
This looks good to me.  One question is with regards to older versions of the command.  Does --no-hints have a long history of support?  Or, if specified on a version where it's not recognized will it just be a no-op or will the program error out in option parsing?

caolanm:
It's not new anyway. already present in version 0.23 as found in RHEL-7.9 for example

Issue #548, opened 2021-11-04
unicode inspection fails if %prep execution fails


dcantrell:
There are a lot of reasons why %prep in the spec file may fail.  So what unicode should do is try to run that, but failing that just fall back on scanning the text files in the SRPM and manually unpacking any archives found in SRPM and scanning those.

PR #551, opened 2021-11-09
[fix] Handle unexecutable %prep sections in 'unicode'

dcantrell:
The unicode inspection looks at source code in the SRPM.  It does this
by setting things up to run through %prep in the same way that
rpmbuild would.  The problem is that rpminspect does not enforce
BuildRequires or other build system specifics, so %prep execution may
fail.  In cases where it fails, have the unicode inspection fall back
on manually extracting any source archives it finds in the SRPM and
inspecting all of those.

The inspection already looks at all text files in the SRPM, so this
will ensure we get reasonable coverage even when the source tree
cannot be set up the same way as %prep would do it.

Fixes: #548

### DIFF
--- a/lib/inspect_unicode.c
+++ b/lib/inspect_unicode.c
@@ -25,6 +25,9 @@
 #include <errno.h>
 #include <err.h>
 #include <ftw.h>
+#include <sys/types.h>
+#include <sys/wait.h>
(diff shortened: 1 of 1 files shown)


jimbair:
Looks good to me!

PR #552, opened 2021-11-10
[fix] Remove hardcoded maximum version check on Java bytecode

gnu-andrew:
The maximum version check of 60 for Java bytecode now fails when
encountering valid bytecode from Java 17 (61) or Java 18 (62).  As
Java releases now occur on a six-month cycle, retaining the current
version check would mean bumping this line a couple of times a year.

It seems better to just remove the check as a maximum bound check is
already performed against supported_major, which can be configured
externally without rebuilding rpminspect.

### DIFF
--- a/lib/inspect_javabytecode.c
+++ b/lib/inspect_javabytecode.c
@@ -119,7 +119,7 @@ static bool check_class_file(struct rpminspect *ri, const char *fullpath,
     /* basic checks on the most recent build */
     if (major == -1 && !strsuffix(localpath, CLASS_FILENAME_EXTENSION)) {
         return true;
-    } else if (major < 0 || major > 60) {
+    } else if (major < 0) {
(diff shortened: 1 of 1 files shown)


dcantrell:
Looks good to me.  Thanks for the patch.

For the history, that comparison comes from rpminspect's ancestor and I lifted the comparison where it checked to see if the major version was less than 0 or greater than 60.  The comment denotes "jdk 15" and the error message reported says "java byte code version is incorrect, wrong endien? corrupted file? space JDK?" (with the misspelling on endian).  I took this to indicate a special case that I should be accounting for, but I will definitely defer to your knowledge here.

Thanks!

gnu-andrew:
Thanks for the quick review.

On the subject of rpmdiff, what's the process for resolving the same issue there? java-17-openjdk is in RHEL 8.5 and produces these same failures. At the moment, we have an autowaiver in place, but that's not sufficient as the sheer number of these warnings generated causes rpmdiff to reach its limit.

PR #554, opened 2021-11-11
Create SimpleSrpmBuild class in the test suite and update tests

dcantrell:
rpmfluff provides SimpleRpmBuild which the test suite uses extensively.  But there are some test cases that just need to create an SRPM, or more importantly a bad SRPM.  I created the SimpleSrpmBuild class that inherits from SimpleRpmBuild but only creates an SRPM for use in the test cases.  This adjusts the class inheritance in baseclass.py but everything should still work as expected.  The SRPM-specific test cases throughout the test suite will now just have an SRPM built instead of all of the packages.


jimbair:
Looks good to me

PR #556, opened 2021-11-11
[tst] Add unicode test cases covering rpmSpecBuild() failures

dcantrell:
In cases where rpminspect cannot run the %prep section of the spec
file, it will fall back on manually unpacking any source archives
found and scanning all of the files in there.


jimbair:
LGTM =)

Issue #583, opened 2021-12-14
no fallback to 'GENERIC' icon list of vendor data doesn't provide one


caolanm:
So the attached example fails with "1) Desktop file /usr/share/applications/libreoffice-startcenter.desktop on x86_64 references icon document-new but no subpackages contain document-new" on fedora/centos/rhel configurations but is ok with --release='GENERIC'
[desktopfilestest-1-1.fc35.src.rpm.zip](https://github.com/rpminspect/rpminspect/files/7710412/desktopfilestest-1-1.fc35.src.rpm.zip)

dcantrell:
This was happening because I neglected to install the /usr/share/rpminspect/icons/* files in the rpminspect data package.  I noted in the internal bug report that I fixed that in rpminspect-data-redhat and a new build kicked off.

I modified the desktop inspection to get a list of standard icon names from /usr/share/rpminspect/icons/PRODUCT where PRODUCT matches the product release string rpminspect either derives on its own or you specify on the command line.  For example, Fedora 35 would be "fc35".

I noticed the same problem I made in the rpminspect-data-fedora and rpminspect-data-centos packages, so I fixed those too.  New Copr builds are available and the next release I do of the data packages will contain the fix as a stable build.

dcantrell:
Verified this is now passing with the latest rpminspect and corresponding data package.

PR #584, opened 2021-12-14
fallback to 'GENERIC' icons list if no vendor one found

caolanm:
https://github.com/rpminspect/rpminspect/issues/583

### DIFF
--- a/lib/init.c
+++ b/lib/init.c
@@ -1889,6 +1889,12 @@ bool init_icons(struct rpminspect *ri)
     assert(ri->icons_filename != NULL);
     contents = read_file(ri->icons_filename);
 
+    /* fallback to 'GENERIC' if not found */
+    if (contents == NULL && strcmp(ri->product_release, "GENERIC")) {
+        xasprintf(&ri->icons_filename, "%s/%s/%s", ri->vendor_data_dir, ICONS_DIR, "GENERIC");
+        contents = read_file(ri->icons_filename);
+    }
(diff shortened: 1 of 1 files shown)


dcantrell:
See comments in #583 as well.

I am not really a fan of this approach.  The GENERIC files are really provided in this repo as templates for the creation of actual rpminspect-data packages.  The above will only work if the user also has rpminspect-data-generic installed, which I sort of intended to be treated like a generic-logos package or something similar.  That package is not required by anything, so the user would have to explicitly have it installed.

I am not opposed to some sort of fallback method, but would prefer it instead to try to fall back to a previous release -or- fall back to maybe the most recent mtime'd file in /usr/share/rpminspect/icons, but also note that in the Details output that it fell back to a different file.  If this was done, I would kind of prefer it to be a generic function that could be used for fallback purposes in the other inspections that read data files from /usr/share/rpminspect.

If you want to leave this as an RFE with me, that's perfectly ok.  I'll likely get around to a fallback system in January as I am trying to finish up some other inspections and test cases at the moment.

caolanm:
I'm not invested in the patch, if the better solution was to update the rpminspect-data-FOO packages instead that's fine by me. Lets just close it and avoid the clutter

dcantrell:
Sounds good to me.  I'm going to at least the fallback idea in the back of my head, but there's still plenty for me to work on in the interim.  :)

PR #592, opened 2022-01-23
Implement rpmdeps inspection with test cases

dcantrell:
This large PR implements the 'rpmdeps' inspection which checks for correctly formatted RPM dependencies in each RPM header.  Things like ensuring that a subpackage that has an automatic shared library dependency on another subpackage carries an explicit Requires on the subpackage providing the shared library.  Or making sure if the package sets an Epoch that the explicit dependencies in subpackages also carry the Epoch value in the V-R expression so it would be E:V-R.  Or reporting multiple providers of the same shared library.

This inspection also supports RPM weak dependencies when the version of RPM on the system is >= 4.12.0.  Things like Suggests, Supplements, Recommends, and Enhances.  The rules around weak dependencies are less strict than Provides and Requires, but they should still be properly formatted.

The test suite gains 393 unique test cases just for the rpmdeps inspection.  The CI scripts and osdeps files for different distributions needed some additional work to ensure things run correctly.


jimbair:
Looks good to me; this is a very large PR, but nothing obviously wrong here, so I say we merge it and start testing it with a wider audience. :)

jimbair (approved):

PR #610, opened 2022-02-07
[ci] Pull CentOS images from quay.io, prepare for CentOS 9 Stream

dcantrell:
CentOS images are now published on quay.io, so move over to using that
host as the origin.  Requires login credentials, which are stored as
repository secrets in GitHub and then referenced here as variables.

Also start support for building and testing on CentOS 9 Stream.
There's not enough EPEL-9 yet to make that work, so just leave it as a
work in progress.


.github/workflows/centos.yml
```diff
@@ -26,21 +26,38 @@ jobs:
         runs-on: ubuntu-latest
 
         # Set up the matrix of distributions to test
+        #
+        # Once EPEL-9 is filled out more, add c9s to the mix:
```
jimbair, line 30: I'm curious what's missing from EPEL-9 to add stream9 into the pipeline so we can either watch for it going forward, or see if we can poke someone in CPE to add them.

jimbair (approved):

.github/workflows/centos.yml
dcantrell: It's about 6 packages right now.  I have BZs open requesting EPEL-9 branches for some of my packages from other people, so I feel we just need some time to sort this out.  EPEL-9 branches are not automatic because it does rely on an active maintainer so this also serves as a nice way to determine from EPEL people are actually using in the new version.

PR #633, opened 2022-02-17
[lib] Add libxdiff directly to the source tree

dcantrell:
This libxdiff comes from this git project:
    https://github.com/spotrh/libxdiff

This project is available in Fedora, but is not commonly available in
other distributions.  This libxdiff is also from git and libgit2 and
extracted from those projects and turned in to a shared library.  The
previous libxdiff (version 0.23) is API-incompatible and tends to more
available in other distributions than this git/libgit2 libxdiff.  To
make things easier for myself and users, I am just adding this special
libxdiff in to this source tree.


jimbair (approved):

jimbair:
Yeah, given the odd nature it's probably best to bring this along. Maybe in a few years we can reevaluate and go back to using local packages but for now, this should make things easier.

PR #643, opened 2022-02-24
[lib] Pretend that TTY has 80 columns, if the real width is unknown

msrb:
80 columns ought to be enough for anybody.

Returning zero from tty_width() could cause problems down the road
as we could easily end up with negative bar_width in setup_progress_bar().

### DIFF
--- a/lib/tty.c
+++ b/lib/tty.c
@@ -27,15 +27,18 @@
 
 /*
  * Return the terminal width.  Used by output routines sending text to
- * stdout.  This function returns 0 if it cannot figure out the width,
- * which means callers should just not worry about the terminal.
+ * stdout.  This function returns 80 if it cannot figure out the width.
  */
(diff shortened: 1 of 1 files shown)


msrb:
We are hitting a bug in CI where rpminspect prints an infinite number of `#` when running with `--verbose`.

I think this patch could fix the problem, but I haven't tested it yet.

dcantrell (changes requested):
This is fine.  I suppose handling cases where stdout is not a tty is something we could do.  I would prefer you define this as DEFAULT_WIDTH in include/constants.h and then return DEFAULT_WIDTH here.  And change the comments accordingly.  I would also prefer it to "the default width" in the comment rather than "the standard width" because this is a topic that gets a lot of opinions and I want to make sure the code remains neutral.

dcantrell:
Modified and merged as #653

PR #660, opened 2022-03-10
[fix] Handle NULL result->msg in output_xunit()

dcantrell:
The output_xunit() function was not properly handling empty
result->msg strings, which could cause a SIGSEGV.

### DIFF
--- a/lib/output_xunit.c
+++ b/lib/output_xunit.c
@@ -79,36 +79,51 @@ void output_xunit(const results_t *results, const char *dest, const severity_t t
             count = 1;
         }
 
-        if (result->severity >= threshold) {
-            fprintf(fp, "        <failure message=\"%s\">%s</failure>\n", result->msg, inspection_header_to_desc(result->header));
-        }
-
         /* prepare the system out message */
         if (result->msg != NULL) {
+            if (result->severity >= threshold) {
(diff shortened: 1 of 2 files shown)


lib/output_xunit.c
```diff
@@ -79,36 +79,51 @@ void output_xunit(const results_t *results, const char *dest, const severity_t t
...
+            assert(msg != NULL);
             free(rawcdata);
         }
 
         /* escape the string for XML CDATA use */
         cdata = strxmlescape(msg);
+        assert(cdata != NULL);
         fprintf(fp, "        <system-out><![CDATA[%s]]></system-out>\n", cdata);
         free(cdata);
-        free(msg);
```
jimbair, line 111: This is probably my C ignorance, but I see we remove free(msg) but not free(cdata), yet we define both as NULL a little bit later. Should both free() calls be removed?

dcantrell: Good catch.  Let me doublecheck things on my local machine before merging.  As-is this either leaks memory or performs a double free.

dcantrell: OK, valgrind reports malloc trouble in strxmlescape(), which is used by output_xunit() and that may ultimately be the source of why this free(msg) was removed.  Working on strxmlescape() now.

jimbair (approved):

Issue #691, opened 2022-03-25
virus: "ignore" config option doesn't seem to work


msrb:
Copied from https://github.com/rpminspect/rpminspect/issues/683#issuecomment-1078942570

`ignore:` is also behaving strangely for `virus` inspection here:
https://src.fedoraproject.org/fork/msrb/rpms/community-mysql/blob/1e1cdd67e1f1ca9385e63fe4c408347545ecac9e/f/rpminspect.yaml

```shell
$ rpminspect -D 
# rpminspect configuration

---
common:
    workdir: /var/tmp/rpminspect
commands:
    msgunfmt: msgunfmt
    desktop-file-validate: desktop-file-validate
    annocheck: annocheck
    abidiff: abidiff
    kmidiff: kmidiff
vendor:
    vendor_data_dir: /usr/share/rpminspect
    favor_release: newest
inspections:
    license: on
    emptyrpm: on
    lostpayload: on
    metadata: on
    manpage: on
    xml: on
    elf: on
    desktop: on
    disttag: on
    specname: on
    modularity: on
    javabytecode: on
    changedfiles: on
    movedfiles: on
    removedfiles: on
    addedfiles: on
    upstream: on
    ownership: on
    shellsyntax: on
    annocheck: on
    dsodeps: on
    filesize: on
    permissions: on
    capabilities: on
    kmod: on
    arch: on
    subpackages: on
    changelog: on
    pathmigration: on
    lto: on
    symlinks: on
    files: on
    types: on
    abidiff: on
    kmidiff: on
    config: on
    doc: on
    patches: on
    virus: on
    politics: on
    badfuncs: on
    runpath: on
    unicode: on
    rpmdeps: on
desktop:
    desktop_entry_files_dir: /usr/share/applications
ownership:
    bin_paths:
        - /bin
        - /sbin
        - /usr/bin
        - /usr/sbin
    bin_owner: root
    bin_group: root
shellsyntax:
    shells:
        - sh
        - ksh
        - zsh
        - csh
        - tcsh
        - rc
        - bash
filesize:
    size_threshold: info
specname:
    match: full
    primary: name
abidiff:
    suppression_file: .abignore
    debuginfo_path: /usr/lib/debug/
    include_path: /usr/include/
    security_level_threshold: 2
kmidiff:
    suppression_file: .abignore
    debuginfo_path: /usr/lib/debug/
    kernel_filenames:
        - vmlinux
        - vmlinuz
patches:
badfuncs:
[...]

PR #693, opened 2022-03-25
[fix] Honor all per-inspection ignore lists; match path prefix

dcantrell:
This is a large one.  First, a number of inspections were not honored
in the config file if you specified a per-inspection ignore list.
That has been fixed and all applicable inspections that can have a
per-inspection list now have one.  Some memory management fixes are
also present as well as a fix that prevents a reset of config file
settings as subsequent files are read.  Settings are now cummulative.
This may require additional work if we find it's really common for an
end user to need to override a config file block, but by and large the
most common case is appending so that's now how all settings are read
in.

The other change is with ignore path matching.  Previously all
matching was done using glob(3).  This is fine and works well, but it
does not allow a user to specify a single directory to exclude and
have it exclude everything in that directory.  The way glob(3) handles
directories in the path name is explicit.  Basically globbing only
works for path components and not the directory names themselves.
This is in POSIX.  While glibc offers some flexibility around this,
it's not portable.

So taking a hint from how rsync does things, rpminspect will now treat
ignore list entries ending with a slash ('/') as a path prefix for
ignore purposes.  So in your rpminspect.yaml file if you specify:

    ---
    virus:
        ignore:
            - /usr/share/mysql-test/

It will ignore /usr/share/mysql-test and everything in that directory.
Without that, the package maintainer would need to explicitly list
glob(3) patterns for each subdirectory construction in that path
prefix which is tedious to maintain and unnecessary.

The data/generic.yaml template file has been updated to reflect all
inspection blocks that were added to honor a per-inspection ignore
list.

Fixes: #691

### DIFF
--- a/data/generic.yaml
+++ b/data/generic.yaml
@@ -351,6 +351,22 @@ addedfiles:
     #ignore:
     #    - /usr/lib*/libexample.so*
 
+movedfiles:
+    # Optional list of glob(7) specifications to match files to ignore
+    # for this inspection.  The format of this list is the same as the
+    # global 'ignore' list.  The difference is the items specified
+    # here will only be used during this inspection.
+    #ignore:
+    #    - /usr/lib*/libexample.so*
+
+removedfiles:
+    # Optional list of glob(7) specifications to match files to ignore
+    # for this inspection.  The format of this list is the same as the
+    # global 'ignore' list.  The difference is the items specified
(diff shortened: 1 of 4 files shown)


jimbair (approved):

msrb:
Amazing! Thanks David :wink:

Issue #696, opened 2022-03-28
RFE: expand patches inspection to ensure all listed patches are applied


dcantrell:
Make sure all listed patches in the spec file are applied, warn if not.

PR #721, opened 2022-04-22
[lib] Expand the patches inspection to verify patches are applied

dcantrell:
Expand the patches inspection to verify that all defined patches in
the RPM header are applied either with %autosetup, %autopatch, or a
corresponding %patchN macro.  If a defined patch is not found to have
a matching apply macro, it raises a VERIFY warning.

Fixes: #696

### DIFF
--- a/include/constants.h
+++ b/include/constants.h
@@ -616,11 +616,67 @@
 #define SPEC_MACRO_GLOBAL      "%global"
 
 /**
- * @def SPEC_SECTION_CHANGELOG
+ * @def SPEC_MACRO_AUTOPATCH
  *
- * How the change log is specified in the spec file.
+ * The %autopatch macro used to apply all defined Patches.
  */
-#define SPEC_SECTION_CHANGELOG "%changelog"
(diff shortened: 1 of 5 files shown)


include/constants.h
```diff
@@ -616,11 +616,67 @@
...
- * How the change log is specified in the spec file.
+ * The %autopatch macro used to apply all defined Patches.
  */
-#define SPEC_SECTION_CHANGELOG "%changelog"
+#define SPEC_MACRO_AUTOPATCH   "%autopatch"
+
+/**
+ * @def SPEC_MACRO_AUTOSETUP
+ *
+ * The %autosetup macro used to unpack all Sources and apply all
```
jimbair, line 628: Small nit: Sources and Patches capitalized; not sure if we're referencing a "proper" thing or if it's just a rogue shift key here and there.

jimbair (approved):

include/constants.h
dcantrell: The naming is deliberately.  I was trying to make the comments allude to what someone would see in a spec file.  Like a PatchN: or SourceN: header.

Issue #703, opened 2022-04-05
rpminspect may be choosing debuginfo files more or less arbitrarily when doing annocheck test


mulkieran:
See https://osci-jenkins-1.ci.fedoraproject.org/job/fedora-ci/job/rpminspect-pipeline/job/master/90229/testReport/(root)/tests/_annocheck/.

Also, stratisd specfile[1] will tell you what files should be using the stratisd-dracut debuginfo rpm and what should be using the stratisd debuginfo rpm.

As far as I can tell, the build process is creating the debuginfo rpms correctly.

See prior discussion in bz[2].

[1] https://src.fedoraproject.org/rpms/stratisd/blob/rawhide/f/stratisd.spec
[2] https://bugzilla.redhat.com/show_bug.cgi?id=2071463

dcantrell:
It's not arbitrary, but it does assume that the debuginfo construction is left to rpmbuild.  It was only a matter of time before I had an example that demonstrated something else.  I'll get on this after a couple other pending fixes.  Thanks.

dcantrell:
PR that addresses the debuginfo arg problem.  That appears to be correct now, though annocheck is at least failing for me but I am going to leave that to either annocheck expected behavior or a bug.

PR #731, opened 2022-05-03
[lib] Change how debuginfo dirs are matched for files

dcantrell:
The previous debuginfo path finding code relied on the common practice
of the debuginfo packages being named after their corresponding
subpackage name.  For example if you have a package named "tuna" with
subpackages "tuna-libs" and "tuna-devel" then you will probably find
debuginfo packages named "tuna-debuginfo", "tuna-libs-debuginfo", and
"tuna-devel-debuginfo" (subject to the contents of the subpackages).
The matching code simply looked for a debuginfo package name that
matched this naming style.  But that disregards any spec file that
changes where debuginfo files land.

This commit changes how librpminspect finds debuginfo subpackages by
looking for the actual file in question based on rpm's naming pattern.
My first approach was to match based on Build-ID from the ELF
metadata, but that failed a number of test cases I had because of
using older binutils releases, older versions of rpm, or entirely
different language toolchains.  While I am not a fan of matching this
way, it is more correct than what librpminspect was doing.

Fixes: #703

### DIFF
--- a/include/rpminspect.h
+++ b/include/rpminspect.h
@@ -386,12 +386,12 @@ char *bytes_to_str(unsigned char *array, size_t len);
  * IMPORTANT: Do not free the returned string.
  *
  * @param ri The struct rpminspect for the program.
+ * @param file The file we are looking for debuginfo for.
  * @param binarch The required debuginfo architecture.
- * @param subpkg Optional subpackage name to get the debuginfo for.
  * @return Full path to the extract before build debuginfo package, or
  * NULL if not found.
  */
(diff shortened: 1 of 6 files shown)


include/rpminspect.h
```diff
@@ -401,12 +401,13 @@ const char *get_before_debuginfo_path(struct rpminspect *ri, const char *binarch
  * IMPORTANT: Do not free the returned string.
  *
  * @param ri The struct rpminspect for the program.
+ * @param file The file we are looking for debuginfo for.
```
jimbair, line 404: This could probably be re-worded a bit; maybe something like

"The file that debuginfo should be associated with"

Or something along those lines.

PR #746, opened 2022-05-11
[fix] Relax the 'types' inspection a bit

dcantrell:
The types inspection is intended to catch things like /usr/bin/program
was an executable, but did you know it's now a JPEG file in this
build?  But the libmagic database changes frequently enough for this
to be a bit pedantic.  For example, a recent build of the kernel had
this inspection failing because Makefiles now show up as
"text/x-makefile" instead of "text/plain".  So the magic database got
a little more finegrained.  That's fine, but it should not fail
rpminspect.

What the types inspection does now is break apart the MIME type at the
'/' and compares the first part and the second part.  It's only a
failure if both parts fail to match.  So since 'text' is present in
both of these, it's not a failure.  But it will report something like
text/plain changing to application/x-executable.

### DIFF
--- a/lib/inspect_types.c
+++ b/lib/inspect_types.c
@@ -29,8 +29,14 @@ static bool result = true;
 static bool types_driver(struct rpminspect *ri, rpmfile_entry_t *file)
 {
     const char *arch = NULL;
-    char *bt = NULL;
-    char *at = NULL;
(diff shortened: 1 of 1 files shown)


jimbair (approved):
I like this change as it aims at the heart of the inspection without blocking unnecessarily

PR #780, opened 2022-06-08
[fix] Correct results reporting for the filesize inspection

dcantrell:

### DIFF
--- a/lib/inspect_filesize.c
+++ b/lib/inspect_filesize.c
@@ -28,6 +28,8 @@
 
 #include "rpminspect.h"
 
+static bool reported = false;
+
 static bool filesize_driver(struct rpminspect *ri, rpmfile_entry_t *file)
 {
     bool result = true;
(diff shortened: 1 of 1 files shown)


lib/inspect_filesize.c
```diff
@@ -125,12 +127,14 @@ static bool filesize_driver(struct rpminspect *ri, rpmfile_entry_t *file)
         params.severity = RESULT_INFO;
         params.verb = VERB_OK;
         params.remedy = NULL;
+        result = true;
```
jimbair, line 130: Should this be result or reported? I see:

`bool result = true;`

up on line 35

dcantrell: It needs to be result.  reported is set to true after I call add_result()

jimbair: Thanks for the confirmation!

jimbair (approved):

PR #791, opened 2022-06-14
[fix] Allow NULL inputs to strprefix() and strsuffix()

dcantrell:
Allow NULL inputs to strprefix() and strsuffix().  Just return false
in these cases rather than crashing.


jimbair (commented):
Looks good to me

PR #805, opened 2022-06-23
[fix] Support per-file allowed lists for the badfuncs inspection

dcantrell:
Some programs and libraries deliberately use functions that have been
deemed forbidden by policy.  An example would be a library providing
runtime support for older software.  In cases like this, you can
specify an allowed block listing path globs and the specific forbidden
functions they are allowed to use.  It is adviseable to add a comment
idnicating why the forbidden function is allowed so future readers
will understand why the rpminspect.yaml file is configured that way.

The path specifications can be explicit paths (local path notation,
relative to /) or patterns that are compatible with fnmatch(3) and
glob(3).

For example:

    ---
    badfuncs:
        allowed:
            /usr/sbin/somenetcmd:
                - inet_aton
                - gethostbyname
            /usr/lib*/libsomething.so.*:
                - inet_network

You can place this block in the package's rpminspect.yaml file to
further control how the badfuncs inspection works for your build.

Fixes: #573

### DIFF
--- a/data/generic.yaml
+++ b/data/generic.yaml
@@ -759,6 +759,21 @@ badfuncs:
     - rexec
     - rresvport
 
+    # Some libraries deliberately use forbidden functions.  In these
+    # cases, you can specify an 'allowed' block listing path globs and
+    # the specific functions they are allowed to use.  It is
+    # adviseable to add a comment indicating why the function is
+    # allowed to use forbidden functions for future reference.  The
+    # path specifications can be explicit paths or patterns that are
+    # compatible with fnmatch(3) and glob(3).
+    #
+    #allowed:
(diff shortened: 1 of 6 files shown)


jimbair (approved):

hroncok:
Thank you. I'm finally trying this in https://src.fedoraproject.org/rpms/python3.11/pull-request/65 -- hoping the CI installs rpminspect from your copr.

hroncok:
It says "rpminspect version 1.10" but badfuncs fails. Am I doing it wrong?

hroncok:
> It says "rpminspect version 1.10" but badfuncs fails. Am I doing it wrong?

I was doing it wrong. Trying again.

hroncok:
Same resuts.

Issue #802, opened 2022-06-23
Removed sources reported as removed patches


hroncok:
In here: https://src.fedoraproject.org/rpms/python-mistune/pull-request/6 Zuul reports:

 * [result.json](https://github.com/rpminspect/rpminspect/files/8965725/result.json.txt)

Most importantly the following:

```json
  "upstream": [
    {
      "result": "VERIFY",
      "waiver authorization": "Anyone",
      "message": "New upstream source file `mistune-0.8.4.tar.gz` appeared",
      "remedy": "Unexpected changed source archive content. The version of the package did not change between builds, but the source archive content did. This may be deliberate, but needs inspection. If this change is expected, update \/usr\/share\/rpminspect\/rebaseable\/.fc37 and send a patch to the project that owns that file."
    },
    {
      "result": "VERIFY",
      "waiver authorization": "Anyone",
      "message": "Source file `v0.8.4.tar.gz` removed",
      "remedy": "Unexpected changed source archive content. The version of the package did not change between builds, but the source archive content did. This may be deliberate, but needs inspection. If this change is expected, update \/usr\/share\/rpminspect\/rebaseable\/.fc37 and send a patch to the project that owns that file."
    }
  ]
```

This is expected, the source tarball was changed, but the version was not changed, we should VERIFY. All good.

But:

```json
  "patches": [
    {
      "result": "INFO",
      "waiver authorization": "Not Waivable",
      "message": "Patch file `v0.8.4.tar.gz` removed"
    }
  ]
```

This is most likely bogus. `v0.8.4.tar.gz` was not a patch file.

dcantrell:
I am able to recreate this locally, but the fix is going to be more than a quick simple thing.  I am slating it for v1.11.  Currently trying to button things up for v1.10.

dcantrell:
I can't find build python-mistune-0.8.4-6.fc37 anymore.  What two builds can I use that will let me recreate this issue locally?

hroncok:
You can scratch build from that PR and from the branch just before that PR.

hroncok:
I've also retriggered zuul on the PR to get a fresh scratch build.

PR #836, opened 2022-07-18
[fix] Report removed patches by checking RPMTAG_PATCHES list

dcantrell:
The patches inspection performs a number of checks on patch files in
SRPMs.  The last thing it does is report any removed patches when
comparing two builds.  However, this removal check was done checking
the entire list of files in the SRPM rather than just looking at the
RPMTAG_PATCH files from the SRPM header.  This patch corrects that
error.

Fixes: #802


jimbair (approved):

hroncok:
Thank you.

Issue #799, opened 2022-06-22
MPI suffixes in shared library rpmdeps are not handled correctly, giving bogus errors


jwakely:
See https://fedora.softwarefactory-project.io/rpminspect-report/?url=https://centos.softwarefactory-project.io/logs//2/2/e3e0e446eddd9ae6fc21a11e533f535fee1a8b6e/check/rpm-rpminspect/d588e2a/result.json where we have **rpmdeps** results like:

> Subpackage boost-graph-mpich on x86_64 carries 'Requires: libboost_mpi.so.1.75.0()(64bit)(mpich-x86_64)' which comes from subpackage boost-openmpi but does not carry an explicit package version requirement. Please add 'Requires: boost-openmpi = %{version}-%{release}' to the spec file to avoid the need to test interoperability between various combinations of old and new subpackages.
>
>    Remedy: Add the indicated explicit Requires to the spec file for the named subpackage. Subpackages depending on shared libraries in another subpackage must carry an explicit 'Requires: SUBPACKAGE_NAME = %{version}-%{release}' in the spec file.

This is incorrect, the `boost-openmpi` package does **not** provide that, it provides `libboost_mpi.so.1.75.0()(64bit)(openmpi-x86_64)`, note the different suffix: (openmpi-x86_64) vs (mpich-x86_64)

It seems that rpminspect is ignoring these additional MPI suffixes and then giving bogus errors and suggesting bogus remedies.

The next **rpmdeps** error is:

> Multiple subpackages provide 'Requires: libboost_mpi.so.1.75.0()(64bit)(mpich-x86_64)': boost-mpich, boost-openmpi
>
>    Remedy: Check subpackage %files sections and explicit Provides statements. Only one subpackage should provide a given shared library. Shared library names are automatically added as Provides, so there is no need to specify them in the spec file but you do need to make sure only one subpackage is packaging up the shared library in question.

This is also bogus, for exactly the same reason. `boost-mpich` provides that lib with a (mpich-x86_64) suffix and boost-openmpi provides it with a (openmpi-x86_64) suffix. That's not the same thing.

The other seven rpmdeps errors in that report are due to the same issue.

dcantrell:
The trailing qualifiers like (x86-64) and (mpich-x86_64) should be excluded for the purposes of comparisons at that point, but I obviously have a mistake here.  Digging in to it.

dcantrell:
[...]

PR #837, opened 2022-07-19
[fix] Do not remove all () substrings from requirements in rpmdeps

dcantrell:
In the rpmdeps inspection, a check is performed to ensure subpackages
with automatic shared library dependencies carry an explicit named
dependency on the subpackage providing that shared library.  Part of
that comparison trims the ISA substring that appears in the
requirement strings, but that was making the assumption that all
substrings in parens were ISA substrings.  That's not the case.  This
patch takes care to leave any paren substrings in place and only
remove the ISA substrings.

The ISA substrings are matched using a regular expression.  This is
done to avoid listing out all possible ones as defined in
/usr/lib/rpm/platform/*/macros since that may change from time to
time.

Fixes: #799

### DIFF
--- a/lib/inspect_rpmdeps.c
+++ b/lib/inspect_rpmdeps.c
@@ -19,8 +19,11 @@
  * SPDX-License-Identifier: LGPL-3.0-or-later
  */
 
+#include <stdlib.h>
+#include <string.h>
 #include <assert.h>
 #include <err.h>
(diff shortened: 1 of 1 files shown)


pr-watson (commented):
LGTM

jimbair (approved):

Issue #819, opened 2022-07-05
rpminspect: *** Unable to determine product release or none specified (-r).


jimbair:
While running against some el8 builds, I ran across these two packages not being able to find their product release:

```
[jbair@awvr ~]$ rpminspect-redhat -s VERIFY -w mytmp -o logs/libmodulemd-2.13.0-1.el8.log libmodulemd-2.13.0-1.el8 
rpminspect: *** Unable to determine product release or none specified (-r).
[jbair@awvr ~]$ rpminspect-redhat -s VERIFY -w mytmp -o logs/subscription-manager-migration-data-2.0.51-1.log subscription-manager-migration-data-2.0.51-1 
rpminspect: *** Unable to determine product release or none specified (-r).
[jbair@awvr ~]$ 
```

I'm not sure if something changed on our side as this is a new one from the tag-runner side.

jimbair:
Would this mean we have something like `rpminspect-centos9` vs `rpminspect-centos`? I suspect devs would sooner put in a dist tag vs a yaml config. Also, libmodulemd-2.13.0-1.el8 appears to have el8 in the name; was it also missing the dist tag?

dcantrell:
No, there wouldn't be a new wrapper script.  What this would do is mean the rpminspect.yaml file you have in the RHEL 8.6 branch, for instance, would say "product_release: el8".  Since the runner will always pick up any rpminspect.yaml present, that's how rpminspect could know what product release to use when it can't infer it from the dist tag.  The user still runs the same thing, which would be ```rpminspect-VENDOR [opts]```

jimbair:
This works; I'm not sure how many devs will use it in lieu of a dist tag, but tag-runner could use it as a part of the profile, which would solve missing dist tags in runs. Though even a --product-release flag in the CLI would also work, but in this case I think I could simply generate a yaml file in the CWD and fire off the tag-runner runs.

It's an edge case so I'm not overly worried about it.

PR #838, opened 2022-07-19
[lib] Support a new config file block called 'environment'

dcantrell:
In addition to letting users specify the product release string on the
command line, also introduce a block in the config file where you can
do the same thing.  It would look like this:

    environment:
        product_release: el8

Use this in a per-package rpminspect.yaml file where rpminspect is
unable to detect the product release string from the dist tag.  Some
packages lack a dist tag.

Fixes: #819

### DIFF
--- a/lib/init.c
+++ b/lib/init.c
@@ -1085,7 +1085,7 @@ static int read_cfgfile(struct rpminspect *ri, const char *filename)
                             free(ri->profiledir);
                             ri->profiledir = strdup(t);
                         }
-                    } else if (block == BLOCK_COMMON) {
+                    } else if (block == BLOCK_ENVIRONMENT) {
                         if (!strcmp(key, "product_release")) {
                             free(ri->product_release);
                             ri->product_release = strdup(t);
--- a/lib/init.c
+++ b/lib/init.c
@@ -146,7 +146,8 @@ enum {
     BLOCK_REMOVEDFILES = 68,
     BLOCK_SYMLINKS = 69,
     BLOCK_UPSTREAM = 70,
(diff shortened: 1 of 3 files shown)


lib/init.c
```diff
@@ -1081,6 +1085,11 @@ static int read_cfgfile(struct rpminspect *ri, const char *filename)
                             free(ri->profiledir);
                             ri->profiledir = strdup(t);
                         }
+                    } else if (block == BLOCK_COMMON) {
```
jimbair, line 1088: It looks like BLOCK_COMMON was copied over from the above if statement and needs to be changed to BLOCK_ENVIRONMENT per line 691

src/rpminspect.c
```diff
@@ -606,6 +605,13 @@ int main(int argc, char **argv)
 
     free(profile);
 
+    /* Product release specified on the command line overrides config file */
+    if (release) {
+        free(ri->product_release);
+        ri->product_release = release;
+        free(release);
```
pr-watson, line 612: freeing release here will cause a dangling pointer, improper usage

pr-watson (changes requested):

lib/init.c
dcantrell: Fixed in followup commit, thanks!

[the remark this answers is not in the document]
dcantrell: Fixed in followup commit, thanks!

pr-watson (approved):
LGTM

src/rpminspect.c
dcantrell: Fixed in followup commit, thanks!

Issue #750, opened 2022-05-13
abidiff suppressions should load an SRPM in all cases and per-RPM ones individually


dcantrell:
Per the abipkgdiff instructions, rpminspect should be picking up .abignore files based on filename suffix, but also if they are present in the built RPM and not just the SRPM.  Seems what I saw in rpminspect's ancestor was a simplified method by which we maintained RPMs, but rpminspect should be more adaptable to where .abignore files will show up.

PR #839, opened 2022-07-20
[fix] Correctly build the abidiff(1) command line

dcantrell:
The abidiff(1) command line was incorrectly constructed.  It worked
for smaller packages where there tends to be a single debuginfo
package, but not larger builds with many subpackages.  Here's what has
changed:

* The debuginfo paths are collected and grouped per arch for the
  before and after builds.

* When the abidiff command line is constructed in the abidiff
  inspection, it selects the debuginfo path that is likely to contain
  the debugging symbols for the file in questions.  By likely I mean
  that I use a strstr() match and run with that.  There will probably
  be edge cases that show up later that I will need to account for.

* Dropped all of the headers-dir handling for abidiff(1) because
  that's impossible to determine.  I can only specify a single
  argument to abidiff(1), but it's not possible to figure out which
  tree will contain the necessary header file for the library in
  question since that's not 1:1 file-based matching.  So just stop
  pretending and don't do this.

The abidiff inspection should now be more stable as a result.

Fixes: #750

### DIFF
--- a/data/generic.yaml
+++ b/data/generic.yaml
@@ -674,9 +674,6 @@ abidiff:
     # Location where debuginfo files live in RPM packages.
     debuginfo_path: /usr/lib/debug/
 
-    # Location of header files in RPM packages.
-    include_path: /usr/include/
-
     # Extra arguments for the abidiff(1) program.  These are put on
     # the command line before the two files being compared.
     #extra_args:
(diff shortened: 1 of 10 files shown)


pr-watson (approved):
LGTM

jimbair (approved):

PR #840, opened 2022-07-21
[fix] Make sure abidiff can find all debuginfo type files

dcantrell:
This is just simplifying things more and ensuring abidiff finds all
debugging data for its reporting.

The man page for abidiff(1) indicates multiple --d1 and --d2 arguments
can be specified, which I initially did, but they have to point
directly to the top of the debuginfo path.

### DIFF
--- a/include/constants.h
+++ b/include/constants.h
@@ -305,7 +305,7 @@
  * this is a path to the corresponding directory in a debuginfo
  * package.
  */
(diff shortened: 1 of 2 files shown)


jimbair (approved):

pr-watson (approved):
LGTM

PR #842, opened 2022-07-21
[lib] Add headers dir support back to the the abidiff inspection

dcantrell:
The previous method was not working, so modify things to pass the
correct paths to abidiff(1).  Now the inspection passes debuginfo
paths and header paths.

There may be some issues regarding private headers, but I'm not aware
of any package out there that ships private headers that are installed
on the system.  But I could be wrong.

### DIFF
--- a/include/constants.h
+++ b/include/constants.h
@@ -318,6 +318,22 @@
 #define ABI_DEBUG_INFO_DIR2 "--d2"
 
 /**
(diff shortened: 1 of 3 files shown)


jimbair (commented):
github won't let me "approve" once merged (and I was late to the party), but consider it approved =)

PR #851, opened 2022-07-28
[tst] added integration test for removedfiles inspection

pr-watson:
Signed-off-by: prwatson <prwatson@redhat.com>

### DIFF
--- a/test/test_removedfiles.py
+++ b/test/test_removedfiles.py
@@ -1,0 +1,117 @@
+#
+# Copyright © 2022 Red Hat, Inc.
+# Author(s): David Cantrell <dcantrell@redhat.com>
+#            Preston Watson <prwatson@redhat.com>
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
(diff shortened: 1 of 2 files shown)


dcantrell (changes requested):
Couple of things to do:

- There's no need to duplicate the tests for all of the compressed file types.  Those tests (wherever they were) were meant to test decompression support in rpminspect.  It's fine to just use the first block, like gzip, and forget the rest.
- Likewise, it's not necessary to use the gzip sample files here.  The removedfiles inspection just needs a file, so I usually just use the rpminspect executable or a text file.  Look at some of the other inspections for an example.  But it's fine to use the gzip files too, it just may be confusing to future readers.
- Drop the 'Gzip' prefix on the class names since that's not what these tests are about.
- You need to run 'make flake8' and 'make black' before submitting a PR for the test suite.  Those need to pass and if they don't, you need to reformat the Python code.
- Also, make sure the copyright and Authors list is correct at the top of the file.

test/test_removedfiles.py
```diff
@@ -0,0 +1,286 @@
+#
+# Copyright © 2020 Red Hat, Inc.
+# Author(s): David Cantrell <dcantrell@redhat.com>
```
jimbair, line 3: You'll definitely want to update the copyright and author(s) section to reflect 2022 and you =)

pr-watson: Adding that in the upcoming PR thanks

pr-watson:
Added test_removedfiles.c containing test cases for TestCompareRPMS, TestCompareKoji, and security files. The tests check that the result and waiver_auth are as expected when a file is removed and not removed.

dcantrell:
This PR includes a conflicting change that was already committed as 20ba55597c90d1bf335e4ad0fd54b6c9ecf7c229.  Please remove it.

dcantrell:
PR#887 replaces this PR.

PR #853, opened 2022-08-02
[lib] When downloading modules, only display Downloading msg once

dcantrell:
This is a side effect of the way rpminspect stores the module
information internally.  Regular Koji builds have one buildentry, but
modules have any number of them.  The download loop iterates over the
build entries and for each one displays the "Downloading..." message.
For regular Koji builds this just appears once, but for modules it
appears for each buildentry.  Modify the loop so it only displays once
for module downloads.


jimbair (approved):
Makes sense. I would want to test this a bit to see how it works (does this show a status bar then appear to freeze while 20+ remaining things are fetched?), but if it's too busy (a wall of fast fetches) that's not helpful either.

Issue #859, opened 2022-08-29
src package reported as s390x


hroncok:
When I download the src package from https://koji.fedoraproject.org/koji/taskinfo?taskID=91170165 and run:

```
$ rpminspect-fedora -T license .../python3.11-3.11.0~rc1-2.fc38.src.rpm
```

With the workaround described in https://github.com/rpminspect/rpminspect/issues/858, it reports:

```
1) Valid License Tag in python3.11-3.11.0~rc1-2.fc38.s390x: Python-2.0.1 
```

But this is not s390x, This is src.

<details>

<summary>Full output</summary>

```
$ rpm -qa | grep -E 'rpminspect|license-data'
fedora-license-data-1.3-1.20220825152831108795.main.0.g32b8e63.fc35.noarch
librpminspect-1.10-1.fc35.x86_64
rpminspect-1.10-1.fc35.x86_64
rpminspect-data-fedora-1.9-1.fc35.noarch

$ rpminspect-fedora -T license .../python3.11-3.11.0~rc1-2.fc38.src.rpm 
diagnostics:
------------
1) Version information for libraries and programs used by rpminspect.  This result is for informational and diagnostic purposes only. 

Result: INFO
Waiver Authorization: Not Waivable

Details:
rpminspect version 1.9
zlib version 1.2.11
libmagic version 540
libclamav version 0.103.7
librpm version 4.17.1
libxml version 20914
json-c version 0.15
libcurl version 7.79.1 (OpenSSL/1.1.1q-fips zlib/1.2.11 brotli/1.0.9 libidn2/2.3.3 libpsl/0.21.1 (+libidn2/2.3.2) libssh/0.9.6/openssl/zlib nghttp2/1.45.1 OpenLDAP/2.4.59)
libarchive version 3.5.3 (zlib/1.2.11 liblzma/5.2.5 bz2lib/1.0.8 liblz4/1.9.3 libzstd/1.5.2)
libyaml version 0.2.5
OpenSSL version 1.1.1q  FIPS 5 Jul 2022
xmlrpc-c version 1.51.0
msgunfmt (GNU gettext-tools) 0.21
diffstat version 1.64
annocheck version 10.79.
abidiff version 2.0.0
kmidiff version 2.0.0


2) Command line arguments used to invoke rpminspect. 

Result: INFO
Waiver Authorization: Not Waivable

Details:
[...]

PR #860, opened 2022-08-30
[fix] Handle source packages in get_nevra() (#859)

dcantrell:
Source packages report the rpmbuild runtime architecture when asking
for RPMTAG_NEVRA.  In rpminspect, we want to show "src" for the
architecture of source packages so handle that as a special case.

Fixes: #859

### DIFF
--- a/lib/rpm.c
+++ b/lib/rpm.c
@@ -151,12 +151,21 @@ char *get_nevr(Header hdr)
 }
 
 /*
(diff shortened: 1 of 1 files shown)


jimbair (approved):
As discussed, a NULL in get_rpm_header_arch() is fine if we somehow hit it. Logic follows; looks good to me.

Issue #857, opened 2022-08-29
Make it more obvious how to run rpminspect on local RPM files


hroncok:
I wish to run rpminspect on RPm packages I built with mock or downloaded from a Koji scratch build.

The `--help` only lists this usage:

    rpminspect [OPTIONS] [before build] [after build]

And the man page only lists examples such as:

```
              rpminspect -T ALL -k zlib-1.2.7-1.fc29 zlib-1.2.7-2.fc29

              rpminspect -T license,elfsyms perl-5.28.0-47.fc6 perl-5.28.1-1.fc6

              rpminspect -T !manpage x3270-3.6ga5-6.fc31 x3270-3.6ga6-1.fc31

              rpminspect -T ALL -a ppc64le zsh-5.7.1-3.fc31 zsh-5.7.1-4.fc31

              rpminspect -E disttag -a ppc64le zsh-5.7.1-3.fc31 zsh-5.7.1-4.fc31
```

The man page says:

> Local RPM packages may be specified directly too if you just want to use rpminspect on a single RPM.

How? Where?

> You may specify a single RPM package or two if you want rpminspect to perform the comparison inspections.

What if my build consists of several subpackages?

PR #863, opened 2022-08-31
[doc] Expand build input description in the rpminspect(1) man page

dcantrell:
Explain how to construct local input directories for rpminspect and
how to specify single RPM and source RPM files.

Fixes: #857

### DIFF
--- a/src/rpminspect.1
+++ b/src/rpminspect.1
@@ -1,4 +1,4 @@
-.\" Copyright © 2018 Red Hat, Inc.
+.\" Copyright 2018 Red Hat, Inc.
 .\" Author(s): David Cantrell <dcantrell@redhat.com>
 .\"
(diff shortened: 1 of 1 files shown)


dcantrell:
@hroncok can you have a look at this man page patch.  I tried to expand the explanation for build inputs to rpminspect.

hroncok:
The new line changes make it hard to see what's different. Could you please split that to a separate commit?

hroncok:
Thanks. The man change looks reasonable and I understood what to do, albeit it is quite tedious 
. I'll open a separate RFE to support directly specifying more RPM files and not comparing them.

dcantrell:
Yeah, that's a reasonable RFE.  The mode of operation is based on what rpmdiff did, but is certainly not the only mode of operation that should be supported.  Thanks.

Issue #864, opened 2022-09-05
error: XML-RPC response too large.  Our limit is 524288 characters.  We got 1473813 characters.


msrb:
It seems like rpminspect is hitting some xmlrpc-c limits when dealing with bigger builds (glibc for example):

```
$ rpminspect-redhat glibc-2.34-28.el9_0.2
rpminspect: XML-RPC Fault: Unable to make sense of XML-RPC response from server.  XML-RPC response too large.  Our limit is 524288 characters.  We got 1473813 characters.  Use XMLRPC_TRACE_XML to see for yourself (-509)
```

This xmlrpc-c function might help us here:
http://xmlrpc-c.sourceforge.net/doc/libxmlrpc.html#limit_set

dcantrell:
Yep, pretty easy to reproduce.  Working on a fix.

PR #867, opened 2022-09-06
[fix] Set XML-RPC size limit to SIZE_MAX

dcantrell:
The default is 512K which has worked up until now.  xmlrpc-c uses a
size_t for this size, so just give it SIZE_MAX and let it run with
that.

Fixes: #864


jimbair (approved):
Makes sense to me =)

PR #868, opened 2022-09-07
[ci] Use vbatts/slackware:latest in GitHub Actions

dcantrell:
Rather than vbatts/slackware:current, use vbatts/slackware:latest which is the latest stable release.  The current branch is not always guaranteed to be working and I just noticed test failures because some of the packages were not built against the version of glibc in the tree.


jimbair (approved):
Simple enough

PR #870, opened 2022-09-07
[lib] Minor fix for the download progress bar alignment

dcantrell:
Off by one for the progress bar message length.  Only noticeable on builds that have really long package names.


jimbair (approved):
Oh I have been waiting for this fix; so simple!

PR #873, opened 2022-09-26
[lib] Change licensdb to a string_list_t in struct rpminspect

dcantrell:
This is an internal library change to allow support for multiple license database files in the rpminspect configuration file.  The license inspection has been modified to work with a list of license databases.


jimbair (approved):
Nothing big to say here; changes seem in line and make sense to me.

PR #876, opened 2022-09-27
[lib] Use section and token macros in init.c

dcantrell:
For consistency and my own sake, use macros across the board for section and token strings in init.c.


jimbair (approved):
I like having these all listed in init.h for easy reference and modification

PR #883, opened 2022-10-04
[tst] Add integration tests for annocheck inspection

zmiklank:
Sorry for taking so long, however here is a proposal for adding the annocheck inspection integration test suite.

Eight integration tests are added that cover the annocheck inspection, testing following use-cases:
    * Single binary RPM  (passing and failing case)
    * Two binary RPMs to compare  (passing and failing case)
    * A single Koji build  (passing and failing case)
    * Two Koji builds to compare  (passing and failing case)

### DIFF
--- a/test/meson.build
+++ b/test/meson.build
@@ -117,6 +117,7 @@ if python.found()
     test_suites = [
         'test_abidiff.py',
         'test_addedfiles.py',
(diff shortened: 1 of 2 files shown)


dcantrell:
Wow, this is great!  Thank you so much!

PR #887, opened 2022-10-06
[tst] added integration test for removedfiles inspection

pr-watson:
Five integration tests are added that cover the removedfiles inspection,
testing following use-cases:
    *  Two binary RPMs to compare (removed and not removed case)
    * A single Security RPM  (removed case)
    * Two Koji builds to compare  (removed and not removed case)
The tests check that the result and waiver_auth are as expected when a file is removed and not removed.

### DIFF
--- a/test/test_removedfiles.py
+++ b/test/test_removedfiles.py
@@ -1,0 +1,117 @@
+#
+# Copyright © 2022 Red Hat, Inc.
+# Author(s): David Cantrell <dcantrell@redhat.com>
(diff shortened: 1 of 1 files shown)


dcantrell:
Looks good, except the commit message needs to conform to doc/git.md

Specifically, the short log prefix.  It's also nice to have a longer description explaining what's in the commit.  See 2dcf0af5fece959eef66073d1b0e4e28a795db62 for an example.

PR #898, opened 2022-10-10
[ci] Disable annocheck tests on Amazon Linux 2

dcantrell:
Amazon Linux 2 is too old to run the annocheck tests at this point, so just skip them.

### DIFF
--- a/osdeps/amzn2/reqs.txt
+++ b/osdeps/amzn2/reqs.txt
@@ -1,4 +1,3 @@
-annobin
 bash
 clamav-data
 clamav-devel


jimbair (approved):
Not knowing the nomenclature, I did a little reading and came across:

https://aws.amazon.com/amazon-linux-2/faqs/

Amazon Linux 2 end of support date (End of Life, or EOL) has been extended by one year from 2023-06-30 to 2024-06-30 to provide customers with ample time to migrate to [Amazon Linux 2022](https://aws.amazon.com/linux/amazon-linux-2022/)

I would argue we focus on AL-2022 if AL-2 is basically in the sunset period.

Issue #897, opened 2022-10-10
per-package override of javabytecode config block does not work


dcantrell:
> Right now packages can provide an rpminspect.yaml that overrides the javabytecode block of the /usr/share/rpminspect/fedora.yaml file, which will give per-package adjustments. If this does not work, it's a bug.

Which version of rpminspect should support the override of the whole configuration block?
I have this rpminspect rpm installed: 1.11-0.1.202210071929gitc1f5750.fc36 but it does not seem to work.

I have used the standard /usr/share/rpminspect/fedora.yaml and rpminspect.yaml which aimed to overwrite it looked like this:

```
javabytecode:
    - fc28: 96
    - fc29: 96
    - fc30: 96
    - fc31: 96
    - fc32: 96
    - fc33: 96
    - fc34: 96
    - fc35: 96
    - fc36: 96
    - fc37: 96
    - default: 16
```

But the result (obtained by `rpminspect-fedora -D`) looked like this:

```
javabytecode:
    - fc28: 52 96
    - fc29: 52 96
    - fc30: 52 96
    - fc31: 52 96
    - fc32: 52 96
    - fc33: 55 96
    - fc34: 55 96
    - fc35: 55 96
    - fc36: 55 96
    - fc37: 55 96
    - default: 43 16
```

The log output of rpminspect run stated following line:

```
Java byte code version 52 greater than supported major version 43 for product release .fc36.
```

In this case I would expect comparing the bytecode version either to 55, specified in the default config, or 96, specified by me. However the default value from default config seemed to be used instead.

dcantrell:
@zmiklank

PR #900, opened 2022-10-12
[fix] When reading config files, allow value overrides

dcantrell:
Some config file blocks need to override previous settings from other config files rather than append.  Expand process_table() to handle this and implement it for 'javabytecode' allowing per-package overrides.

Fixes: #897


jimbair (approved):
Seems simple enough; no objections from me

Issue #901, opened 2022-10-12
rpminspect cannot fetch builds from centos-stream koji


msrb:
`rpminspect -f` fails with strange XML-RPC error when talking to CentOS Stream Koji:

```
$ rpm -q rpminspect
rpminspect-1.11-0.1.202210111809git18619c8.fc37.x86_64

$ rpminspect-centos -f postgresql-jdbc-42.2.18-5.el9
rpminspect: XML-RPC Fault: RPC failed at server.  <class 'ValueError'>: invalid literal for int() with base 10: 'postgresql-jdbc-42.2.18-5.el9' (1)
```

However, Fedora Koji seems fine:
```
$ rpminspect-fedora -f postgresql-jdbc-42.5.0-1.fc38
$ ls
postgresql-jdbc-42.5.0-1.fc38
```

msrb:
Hmm, actually it's not just `-f`:
```
$ rpminspect-centos postgresql-jdbc-42.2.18-5.el9
rpminspect: XML-RPC Fault: RPC failed at server.  <class 'ValueError'>: invalid literal for int() with base 10: 'postgresql-jdbc-42.2.18-5.el9' (1)
```

PR #903, opened 2022-10-12
[fix] Handle Koji XML-RPC failure code 1 when using getTaskInfo

dcantrell:
Fixes: #901


jimbair (approved):
Failure handling is always welcomed

Issue #798, opened 2022-06-22
How to inspect modules locally?


ppisar:
I found that rpminspect has "modularity" test. I supplied a modular package to rpminspect and it reported "Result: INFO". According to rpminspect sources it's because rpminspect does not recognize it is a modular package (buildtype). It only switches to modular build type after retrieving data from Koji. Can this be somehow locally overridden?

So I used "rpminspect -c /usr/share/rpminspect/redhat.yaml -d  php-7.4-8040020220622090235.d7c09045" on a RHEL 9.1 machine and got this error:

~~~~
rpminspect: XML-RPC Fault: Value of type INT supplied where string type was expected. (-501)
~~~~

I tried it from a Fedora 35 system for perl-DBI-1.643-3520220228164016.e12b6f3a and that passed. Is there a bug in handling Brew responses? "brew buildinfo php-7.4-8040020220622090235.d7c09045" works for me on that RHEL system.

Next, on Fedora it worked and reported "Result: OK" for  modularity. Fine. But redownloading everything from Koji takes time. I tried using --keep option to preserve the downloaded data:
~~~~
$ rpminspect -c /usr/share/rpminspect/fedora.yaml -a x86_64 -k -v -T modularity perl-DBI-1.643-3520220228164016.e12b6f3a
Downloading module perl-DBI-1.643-3520220228164016.e12b6f3a
[...]
modularity:
-----------
Result: OK
Waiver Authorization: Anyone
[...]
Keeping working directory: /var/tmp/rpminspect/perl-DBI-1.643.2Z8MYK
~~~~

and then pointed rpminspect to the temporary directory:

~~~~
$ rpminspect -c /usr/share/rpminspect/fedora.yaml -a x86_64 -k -v -T modularity /var/tmp/rpminspect/perl-DBI-1.643.2Z8MYK
rpminspect: Fopen failed for /var/tmp/rpminspect/perl-DBI-1.643.2Z8MYK/after/x86_64/perl-DBI-debuginfo-1.643-7.module_f35+13908+53d84a48.x86_64/usr/lib/debug/.build-id/76/6baf9b1dbc15964c8a42b77cdca3493f4ef334: No such file or directory
rpminspect: mkdir: File exists
Running modularity inspection...     pass
diagnostics:
------------
1) Version information for libraries and programs used by rpminspect.  This result is for informational
and diagnostic purposes only. 

Result: INFO
Waiver Authorization: Not Waivable
~~~~

[...]

PR #907, opened 2022-10-13
Support specifying Koji build type as a command line option

dcantrell:
Normally rpminspect autodetects the Koji build type of the inputs, but in some cases you may want to need to do that explicitly.  The -b or --build-type option on rpminspect now permits that.  The -l option displays supported Koji build types.

Also for the modularity and changelog inspections, when skipping due to the incorrect build type, leave an INFO result message explaining why.

Fixes: #798

### DIFF
--- a/include/rpminspect.h
+++ b/include/rpminspect.h
@@ -53,6 +53,9 @@ extern struct inspect inspections[];
 /* List of all output format types (output.c) */
 extern struct format formats[];
 
+/* List of all build types (koji.c) */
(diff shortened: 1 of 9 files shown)


jimbair (approved):
Looks good. At first I was going to nit buildtype(s) vs build_type(s) but I think buildtype(s) without the underscore reads better in some of the code that comes up later.

Still worth bringing up, but otherwise, I think it's good.

PR #910, opened 2022-10-17
Add the 'debuginfo' inspection and test cases

dcantrell:
In rpminspect's ancestor, this was called the "stripped" inspection.  It ensures built packages have debugging symbols stripped and debugdata sections linking them to the corresponding debuginfo packages.  It also ensures debuginfo packages carry debugging symbols.

### DIFF
--- a/data/generic.yaml
+++ b/data/generic.yaml
@@ -126,6 +126,7 @@ vendor:
     #changedfiles: off
     #changelog: off
     #config: off
+    #debuginfo: off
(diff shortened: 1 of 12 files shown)


jimbair (approved):
Looks good; lots of fresh code so there could be little issues hiding, but it all reads well. I also appreciate the TODO for the test vs leaving it missing; a good reminder to patch that up when we can.

PR #911, opened 2022-10-17
[tst] Support KEEP=y when running 'make check'

dcantrell:
To retain intermediate and results files when debugging test cases, support running "make check KEEP=y" to keep those files.  Normally these are removed after each test case is run.


jimbair (approved):

jimbair:
I await the day when someone asks why KEEP=n doesn't work ;) But overall I think it's fineTM

PR #913, opened 2022-10-18
[lib] Refactor code to make better use of list_add()

dcantrell:
Refactor a lot of code to make better use of list_add() where possible.  This is just a function I wrote to handle adding a new 'char *' to a string_list_t.  No reason not to use it, but I think when I wrote list_add() it was after all of the other stuff had been written.


jimbair (approved):
This PR is a solid example of DRY programming principles at work. =) A hearty +1

PR #917, opened 2022-10-20
[fix]: remove leading dot from product release in javabytecode inspection

zmiklank:
The product_release is obtained from dist tag in the rpm's NVR. It contains leading dot. In javabytecode inspection is the dot not needed.

Please feel free to close this, if there exists more appropriate fix or if current behavior is correct and does not need fixing.

### DIFF
--- a/lib/inspect_javabytecode.c
+++ b/lib/inspect_javabytecode.c
@@ -249,8 +249,11 @@ bool inspect_javabytecode(struct rpminspect *ri)
         return false;
     }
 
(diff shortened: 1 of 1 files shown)


dcantrell:
Thanks for the PR.  This is actually a larger problem for the inspections that make use of ri->product_release.  So I would rather modify get_product_release() in src/rpminspect.c to address this for the rest of the program.  I have done that here:

https://github.com/rpminspect/rpminspect/pull/919

PR #922, opened 2022-10-24
[cmd] Trim leading period(s) after product release string matching

dcantrell:
This is a followup to 704521fecfad00f7380a995f6e7749845badaea8.  Do the trimming after the matching rules from the config file are applied.


jimbair (approved):
More deduplication and simplification; I like it

PR #929, opened 2022-10-25
[lib] Change how RPMs are unpacked when peering happens

dcantrell:
Previously rpminspect would unpack each RPM in to its own subdirectory.  Lots of inspections need to examine the files inside an RPM, so this seemed like a reasonable way to do that.  However, some inspections need a view of the entire built package installed.  Things like abidiff want to see the debuginfo and libraries and headers. Depending on what the packager did, those files may all be in one package or multiple subpackages.  To make this easier for inspections that need to see all files extracted, rpminspect now extracts all RPMs for each build to the same root tree (technically one per architecture with src and noarch separated out).  rpminspect still has all of the RPM header information and knows what subpackage a file belongs to, so this is just an on-disk representation to make things easier for different inspections.

### DIFF
--- a/include/constants.h
+++ b/include/constants.h
@@ -135,6 +135,29 @@
 #define DEFAULT_WORKDIR "/var/tmp/rpminspect"
 
 /**
+ * @def ROOT_SUBDIR
+ *
+ * The name of the root subdirectory used in the working directory.
+ * This is where packages are extracted as a simulated root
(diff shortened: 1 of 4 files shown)


jimbair (approved):
Nice work tracking this down and working with the abidiff folks upstream to get this sorted. Small PR with a big impact. =)

PR #931, opened 2022-10-26
[bld] Increase libabigail version dependency to 2.1

dcantrell:
Increase this from 1.8.2 to 2.1 in the spec file.  Newer major version so we should reset the lower bound for what rpminspect runs in Fedora, CentOS, and RHEL environments.


jimbair (approved):
Simple enough

Issue #924, opened 2022-10-25
RFE: show skipped inspections in output


msrb:
Disabled/skipped inspections are currently completely excluded from the rpminspect output. However, for example HTML result viewer could show which inspections were skipped and then people (QE) could easily see that some inspections did not run (because for example package maintainer decided to turn them off in the component-specific config file).

I understand that this is a special case and when people run rpminspect locally, they probably know which inspections will run. Therefore I propose that the skipped inspections could appear in the output only if rpminspect runs with `--verbose`.

This would also probably require introduction of a new result type, maybe `SKIPPED` (?)

cc @dcantrell @jimbair @martinpitt

msrb:
Note it would be great if we could make it easy to distinguish between inspections that were executed and are OK, and inspections that were skipped. Therefore the suggestion to introduce a new result type.

This features was also user-requested in OSCI-2709.

PR #936, opened 2022-10-26
Show skipped inspections in verbose mode and results

dcantrell:
Inspections that are disabled by the config file or by command line option are now reported in both verbose mode output as well as gaining a "SKIP" result entry.  This should make it more obvious that an inspection was disabled when looking at the results.

Fixes: #924


jimbair (approved):
Lots of tiny changes here; hard to tell without running the test suite, but assuming nothing obvious, looks good to me.

PR #943, opened 2022-11-04
[lib] Allow profile specification by full path or basename

dcantrell:
Some users specify a profile as a full path to the file rather than the basename (minus the .yaml extension) in profiledir.  This is understandable considering how the -c option works.  The origin of rpminspect profiles is from early in its creation where I thought it would make sense to have profiles that describe the rpminspect rules for large categories of software.  That really didn't take off as I thought it would, but the support is still there.

This patch allows the -p option to find a profile by basename in the profiledir as well as allowing the user to specify a full path to a profile.

I have also updated the man page to explain how the options work.  In all cases, the -c option is required and rpminspect will fail if you do not provide a configuration file with the -c option.

Fixes: #940

### DIFF
--- a/lib/init.c
+++ b/lib/init.c
@@ -2237,7 +2237,7 @@ struct rpminspect *init_rpminspect(struct rpminspect *ri, const char *cfgfile, c
         ri = calloc_rpminspect(ri);
     }
 
-    /* Read in the config file if we have it */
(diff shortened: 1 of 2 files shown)


jimbair (approved):
This change makes sense to me; I also appreciate the error messages if we hit a failure condition.

PR #944, opened 2022-11-04
Add .mailmap to consolidate contributors and update AUTHORS.md

dcantrell:
This is something I've been meaning to do for a while.  Add a .mailmap file to consolidate committers who have used multiple addresses over time.  Then modify the 'make authors' target to respect that so that I can just type 'make authors' to update the AUTHORS.md file.

### DIFF
--- a/.mailmap
+++ b/.mailmap
@@ -1,0 +1,1 @@
+Serhii Turivnyi <sturivny@redhat.com> sergturi <turivniy@gmail.com>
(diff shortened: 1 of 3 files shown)


jimbair (approved):
Looks good; not sure if it's worth having everyone in there for clarity's sake even if they only have one commit, or maybe even somehow generate the .mailmap file based on commit history? But I've never had to wrangle it before, so I'm 100% spit balling.

dcantrell:
At some point in time I decided to list everyone who contributed.  Lots of projects do that and it felt better than deciding who "did enough" to be listed in the AUTHORS file.  In some cases, people don't want to listed and for that they can just say so and I will leave them out of the file.

PR #945, opened 2022-11-07
[cmd] Add SKIP result type to viewer

martinpitt:
Keep it with a white background to convey emptiness. Draw a light border
around it to make the label actually visible in the legend and details.
    
Sort this before OK, as it's even less severe/important to look at than
that.

---

Fixes the non-styled state at https://msrb.github.io/viewer.html?url=https://msrb.github.io/results.json# . http://localhost:8000/viewer.html?url=https://msrb.github.io/results.json# looks like this now:

image

Of course, let the styling bikeshedding begin! :grin:

### DIFF
--- a/contrib/viewer.html
+++ b/contrib/viewer.html
@@ -45,11 +45,11 @@
         /* define the border here already, to maintain the layout */
         border-width: 1px;
         border-style: solid;
-        border-color: black;
+        border-color: black !important; /* override .result-SKIP */
     }
     .summary-result.selected {
         padding: 10px 2px;
-        border-color: blue;
+        border-color: blue !important; /* override .result-SKIP */
         border-width: 3px;
     }
 
@@ -68,6 +68,7 @@
         text-align: center;
         margin: 2px 0;
     }
+    .result-SKIP { background-color: white; border-width: 1px; border-color: lightgray; border-style: solid; }
     .result-OK { background-color: #E8FFE8; }
     .result-INFO { background-color: lightblue; }
     .result-VERIFY { background-color: darkorange; font-weight: bold; }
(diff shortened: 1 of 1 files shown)


martinpitt:
@msrb FYI

dcantrell:
I prefer the third one.  Or second to last example shown.

martinpitt:
@dcantrell : This isn't a choice -- the first is the "real" screenshot, the other three the screenshots with the respective [Let's get color blind Firefox extension](https://addons.mozilla.org/en-US/firefox/addon/let-s-get-color-blind/) setting.

msrb:
This would be my personal preference:

- move SKIP before OK in the legend
- italic type makes it stand out, but it's just another result type (imo), so I'd go with regular type probably
- color-wise, since I'd move it before OK, I'd go with some gray-ish, or maybe even pure white color (?) Of course only if David finds it easy to distinguish from the rest

These are not strong preferences, so I will be happy either way :)

dcantrell:
Moving it before OK is fine.  If you drop the italics, I would prefer that it be black text with a white background.  SKIP is a result type, but it's the only one where nothing was actually run for the inspection.  Black text with a white background does convey an "emptiness" aspect in the visualizer.

martinpitt:
> italic type makes it stand out,

I mostly did that to disambiguate the result type without colors. It is indeed sufficiently close to "OK" that it doesn't really matter though, and it does draw attention *to* it instead of diverting it elsewhere. :grin: So I'm happy to drop the italics. So I'll make it white again (incidentally, that's what it is now :grin: ) but draw a border around it (otherwise the label looks really broken).

I updated the code and the screenshots in the description. PTAL?

dcantrell:
Looks good to me.  I'm ready to merge.

Issue #803, opened 2022-06-23
Add a test for a static context of a module


ppisar:
I'd like to add a module-level test which checks whether a module's YAML file has /data/static_context value defined to true or not.

Motivation: Static-context modules are handled differently by DNF. They fix maintaining an upgrade path for multi-context stream and changing modular dependencies. However, they are not supported in RHEL 8. RHEL < 8.6 actually reports an an error on them.

See <https://raw.githubusercontent.com/fedora-modularity/libmodulemd/main/yaml_specs/modulemd_stream_v2.yaml> for a YAML specification.

The expected value depends on a distribution:

* RHEL 8 must not have "static_context: true" in the YAML file.
* RHEL ≥ 9 must have "static_context: true: in the YAML file.
* Fedora should ignore this test. A recommended value is "true", but it's not required.

Examples:
* No static context: [perl-DBI-1.643-3520210211145152.2c956793](https://koji.fedoraproject.org/koji/buildinfo?buildID=1708856)
* A static context: [perl-DBI-1.643-3520220228164016.e12b6f3a](https://koji.fedoraproject.org/koji/buildinfo?buildID=1926764)

dcantrell:
Does the static_context value appear under the data section at the first level or does it appear elsewhere (or can it)?

ppisar:
"static_context" appears only in the "data" section at the first level of nesting (i.e. a direct child). That's why I wrote an XPath-like expression "/data/static_context".

PR #947, opened 2022-11-07
[lib] Check /data/static_context in modulemd.txt in 'modularity'

dcantrell:
Expand the modularity inspection to check the /data/static_context setting from the module metadata.  The value is collected and then checked against the configuration file setting.  Reporting is based on what is found vs. the configuration file setting.  For instance, if it is found set in modulemd.txt and the config file says static_context is forbidden, that is a VERIFY result.

Fixes: #803

### DIFF
--- a/data/generic.yaml
+++ b/data/generic.yaml
@@ -259,6 +259,27 @@ metadata:
         - .localdomain
         - .example.com
 
+modularity:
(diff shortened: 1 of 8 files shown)


dcantrell:
Test cases for this are coming, but I wanted to post the PR first.

jimbair (approved):

PR #949, opened 2022-11-08
Allow TestKoji and TestCompareKoji to be base classes; use super().setUp() and super.tearDown()

dcantrell:
This is a refactoring of TestKoji and TestCompareKoji so I can inherit from them to make module base classes.  Also, change all test cases to consistently use super().setUp() and super().tearDown().

### DIFF
--- a/test/baseclass.py
+++ b/test/baseclass.py
@@ -680,6 +680,10 @@ class TestCompareRPMs(RequiresRpminspect):
 
 # Base test case class that tests a fake Koji build
 class TestKoji(TestRPMs):
+    def setUp(self):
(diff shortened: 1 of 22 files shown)


jimbair (approved):
Looks good to me; the only concern is if we accidentally goof a test, though I assume if they're passing, we should be good. But yeah, I think this looks cleaner and consolidates similar baseline configs for tests.

dcantrell:
What was interesting is that by changing TestKoji and TestCompareKoji, it uncovered some mistakes in tests that I was able to fix with the super().setUp() change.  So that's why I did that across the board.

PR #950, opened 2022-11-09
module base classes for the test suite and modularity test cases

dcantrell:
A number of patches here.  It starts with adding base classes for module test cases, then adds test cases, then some fixes for baseclass.py debugging and some actual fixes for the modularity inspection.

### DIFF
--- a/lib/inspect_modularity.c
+++ b/lib/inspect_modularity.c
@@ -185,8 +185,8 @@ static bool check_static_context(struct rpminspect *ri)
                 xasprintf(&params.msg, _("The /data/static_context value in %s matches the value in %s, and the product release rules require the presence of /data/static_context in the module metadata."), ri->before, ri->after);
             } else if (ri->modularity_static_context == STATIC_CONTEXT_RECOMMEND) {
                 xasprintf(&params.msg, _("The /data/static_context value in %s matches the value in %s, and the product release rules recommend the presence of /data/static_context in the module metadata."), ri->before, ri->after);
-            } else {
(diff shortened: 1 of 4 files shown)


lib/inspect_modularity.c
```diff
@@ -222,14 +222,16 @@ static bool check_static_context(struct rpminspect *ri)
...
-            xasprintf(&params.msg, _("The /data/static_context value in %s is %s, and the product release rules do not have a setting for /data/static_context in the module metadata."), ri->after, asc ? _("true") : _("false"));
+        } else if (ri->modularity_static_context == STATIC_CONTEXT_NULL) {
+            xasprintf(&params.msg, _("The /data/static_context value in %s is %s, but the product release rules do not have a setting for /data/static_context in the module metadata."), ri->after, asc ? _("true") : _("false"));
         }
     }
 
     /* report */
-    add_result(ri, &params);
-    free(params.msg);
+    if (params.msg != NULL) {
```
jimbair, line 231: I feel like I've seen similar changes in previous PRs that stripped the explicit != NULL, like this:

`if (params.msg) {`

Looking through a few older PRs, I see this as one example:
https://github.com/rpminspect/rpminspect/pull/947/files#diff-64844a14a84a03825d4995fa21103cf4420cc9f33987286035139c9eceec6678

in lib/builds.c

dcantrell: Yeah, that's a consistency thing.  I tend to not do the explicit '!= NULL' thing.  I'll drop it.

PR #953, opened 2022-11-10
[bld] Add a Makefile target and script to update uthash.h

dcantrell:
uthash.h comes from https://github.com/troydhanson/uthash. Periodically I pull down the latest copy of the header file, but it never really changes that much.  It's nice to check to see if there's major refactoring and things like that.

I can now type "make update-uthash" to get the latest copy in the tree.


jimbair (approved):
Looks good over here

Issue #957, opened 2022-11-15
New patch file `(null)` appeared


hroncok:
**Describe the bug**

When running on Fedora Zuul in https://src.fedoraproject.org/rpms/python-certifi/pull-request/10#comment-121233 a (null) patch was reported:

```json
  "patches": [
    {
      "result": "INFO",
      "waiver authorization": "Not Waivable",
      "message": "New patch file `(null)` appeared"
    },
    {
      "result": "INFO",
      "waiver authorization": "Not Waivable",
      "message": "certifi-2022.9.24-use-system-cert.patch touches 1 file and as many as 106 lines"
    },
    {
      "result": "INFO",
      "waiver authorization": "Not Waivable",
      "message": "Patch file `certifi-2020.11.8-use-system-cert.patch` removed"
    }
  ],
```

**To Reproduce**

```json
  "diagnostics": [
    {
      "result": "DIAGNOSTICS",
      "message": "Version information for libraries and programs used by rpminspect as well as storage requirements.  This result is for informational and diagnostic purposes only.",
      "details": "rpminspect version 1.11\nzlib version 1.2.11\nlibmagic version 541\nlibclamav version 0.103.7\nlibrpm version 4.17.1\nlibxml version 20914\njson-c version 0.15\nlibcurl version 7.82.0 (OpenSSL\/3.0.5 zlib\/1.2.11 brotli\/1.0.9 libidn2\/2.3.3 libpsl\/0.21.1 (+libidn2\/2.3.2) libssh\/0.9.6\/openssl\/zlib nghttp2\/1.46.0 OpenLDAP\/2.6.2)\nlibarchive version 3.5.3 (zlib\/1.2.11 liblzma\/5.2.5 bz2lib\/1.0.8 liblz4\/1.9.3 libzstd\/1.5.2)\nlibyaml version 0.2.5\nOpenSSL version 3.0.5 5 Jul 2022\nxmlrpc-c version 1.51.0\nmsgunfmt (GNU gettext-tools) 0.21\nannocheck version 10.81.\nabidiff version 2.1.0\nkmidiff version 2.1.0\n\nSpace required to download artifacts: 0 bytes (0 B)\nSpace required to unpack artifacts: 355055 bytes (347 KiB)\n"
    },
    {
      "result": "DIAGNOSTICS",
      "message": "Command line arguments used to invoke rpminspect.",
[...]

PR #959, opened 2022-11-15
[fix] Do not report "New patch file (null) has appeared"

dcantrell:
Instead of "(null)", give the actual filename of the new patch.  It's more useful.

Fixes: #957


jimbair (approved):

hroncok:
> It's more useful.

Indeed :D

PR #960, opened 2022-11-16
[ci] Update FreeBSD files in osdeps/freebsd/

dcantrell:
Update the list of packages and Python modules to install and perform some local tests on FreeBSD 13.1.  Still more to do here, but it's moving.


osdeps/freebsd/post.sh
```diff
@@ -1,6 +1,11 @@
 #!/bin/sh
 PATH=/bin:/usr/bin:/sbin:/usr/sbin:/usr/local/bin:/usr/local/sbin
 
+# Install Python modules from ports, but we have to determine the
+# package prefix based on the version of Python installed.
+PKG_PREFIX="py$(python3 --version | cut -d ' ' -f 2 | cut -d '.' -f 1,2 | sed -e 's|\.||g')"
```
jimbair, line 6: This works, though a validation wouldn't go amiss.

PR #961, opened 2022-11-18
More FreeBSD CI patches

dcantrell:
Portability patches and support for options defined in meson_options.txt


jimbair (approved):
Who doesn't love some libcap (and shf)

PR #962, opened 2022-11-23
Move to using the libannocheck library rather than running annocheck(1)

dcantrell:
This is a large patch set.  Also, I forgot to merge the previous PR with the FreeBSD CI fixes so that's bundled in here too.  They are different commits, so I don't care.

libannocheck is also something that can be disabled at build time for systems where libannocheck is not available.  Reporting is similar to what annocheck(1) reports.  A FAIL or MAYB result from a test will trigger an rpminspect failure (a VERIFY result) unless that has been overridden by the rpminspect config file.  The fortify-source check is a security check and will always be RESULT_BAD unless the vendor data package carries a rule instructing rpminspect to report that finding a different way for the file in question.

Test suite updated, meson files updated, spec file updated, and so on.

### DIFF
--- a/lib/inspect_annocheck.c
+++ b/lib/inspect_annocheck.c
@@ -6,81 +6,204 @@
 #include <stdio.h>
 #include <string.h>
 #include <errno.h>
+#include <err.h>
 #include <assert.h>
+#include <libannocheck.h>
 
 #include "rpminspect.h"
 
+/* Global variables */
 static bool reported = false;
 
-/* Trim workdir substrings from a generated string. */
-static char *trim_workdir(const rpmfile_entry_t *file, char *s)
+/*
+ * Try to map the product release string to an appropriate
+ * libannocheck profile.  This is going to be something that will need
+ * maintenance over time.  Since the profile list is embedded in
+ * libannocheck, there's no easy way to discover changes.  The
+ * assumption here is the profile naming scheme remains the same.
+ */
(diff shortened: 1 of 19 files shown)


lib/inspect_annocheck.c
```diff
@@ -6,81 +6,204 @@
...
+    if (annoerr != libannocheck_error_none) {
+         warnx(_("libannocheck_get_known_profiles error: %s"), libannocheck_get_error_message(anno, annoerr));
+         return;
     }
 
-    fl = strlen(file->fullpath);
-    ll = strlen(file->localpath);
+    /* iterate over the profiles to try and find a match */
+    for (i = 0; i < num_profiles; i++) {
+        /* XXX: this needs a map in the config file */
```
jimbair, line 46: Do we need this map before merging?

jimbair (approved):
I see this also purged a XXX further down so I’m guessing it’s more of a future TODO and not a working TODO.

looks good to me :)

lib/inspect_annocheck.c
dcantrell: Good catch.  Forgot to finish that function.

dcantrell:
> I see this also purged a XXX further down so I’m guessing it’s more of a future TODO and not a working TODO.
> 
> looks good to me :)

Yeah, after you pointed that out I finished up that function and pushed another patch to this PR.

Issue #966, opened 2022-12-01
rpminspect incorrectly reports insufficient disk space in /var/tmp/rpminspect


sharkcz:
**Describe the bug**
rpminspect incorrectly reports insufficient disk space in /var/tmp/rpminspect, I suppose this is a big endian related bug.
````
[sharkcz@rock-kvmlp-fedora ~]$ rpminspect-redhat -v -T ALL 49154443  
There is not enough available space to download the requested build.
    Need 38 PiB in /var/tmp/rpminspect, have 82 GiB.
See the `-w' option for specifying an alternate working directory.
rpminspect: unable to download after build: 49154443
rpminspect: Insufficient disk space in the working directory.
````

**To Reproduce**
run `rpminspect-redhat -v -T ALL 49154443 ` on Fedora 36 on s390x

**Expected behavior**
it shouldn't require 38 PiB of free space

**Environment:**
 - Fedora 36 on s390x

This is a regression in `rpminspect-1.11-0.1.202211231559git683d045.fc36.s390x`, because `rpminspect-1.10-1.fc36.s390x` happily downloads the rpms from brew.

sharkcz:
Seems it is not s390x specific, got one on x86 as well
````
[dan@t14s ~]$ rpminspect-redhat -v -k -T ALL 49364862
There is not enough available space to download the requested task.
    Need 16 EiB in /var/tmp/rpminspect, have 39 GiB.
See the `-w' option for specifying an alternate working directory.
rpminspect: unable to download after task: 49364862
rpminspect: Insufficient disk space in the working directory.
````

sharkcz:
still a problem with `rpminspect-1.11-0.1.202212071943git6a8bb58.fc36.s390x`

dcantrell:
This is caused by my use of unsigned long and unsigned long long rather than int32_t and int64_t when decomposing XMLRPC_TYPE_INT and XMLRPC_TYPE_I8 values.  Patch coming shortly.

sharkcz:
I think I see what is the problem here. It is regular vs scratch build. When rpminspect is pointed to a scratch build, it fails with the error above, when it is a regular build, it runs OK.

sharkcz:
`koji download-task` works for scratch builds, so maybe there is a way ... It will download rpms for just one arch (buildArch task), but still could be useful. I have spent quite some time writing various tools using Koji Python API, so I think it might be possible to learn `rpminspect` about scratch builds too. But let's leave it open for the next year :-)

PR #974, opened 2022-12-10
[fix] Read XMLRPC_TYPE_INT and XMLRPC_TYPE_I8 correctly

dcantrell:
When decomposing an XMLRPC_TYPE_INT librpminspect should use an int32_t and when decomposing an XMLRPC_TYPE_I8 librpminspect should use an int64_t.  This ensures the precision indicated in xmlrpc-c is correct given the differences in long int and long long int by platform.

For consistency, the existing XMLRPC_TYPE_INT values read by librpminspect now read in to an int32_t rather than an int.

Fixes: #966

### DIFF
--- a/include/rpminspect.h
+++ b/include/rpminspect.h
@@ -477,7 +477,7 @@ char *get_file_delta(const char *a, const char *b);
  * @param path The filesystem path to check.
  * @return Number of bytes available.
  */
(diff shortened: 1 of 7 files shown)


jimbair (approved):
Makes sense to me

PR #975, opened 2022-12-12
Take rpminspect back to using /usr/bin/annocheck

dcantrell:
Due to some limitations in libannocheck, we need to switch back to using the executable.


jimbair (approved):
I prefer this patch over a full revert, so it can be rolled back to the library method once it's working the way we need in the future.

PR #976, opened 2022-12-12
[ci] Add 'env CRYPTOGRAPHY_DONT_BUILD_RUST=1' to PIP_CMD for fedora

dcantrell:
We don't need the Rust stuff in the Python modules for what we are using them for, so just disable it in our CI jobs.


jimbair (approved):
Rust be gone

PR #979, opened 2022-12-15
[fix] Use libannocheck_get_version() in libannocheck_init() call

dcantrell:
Rather than using the LIBANNOCHECK_VERSION macro, use the function so that at runtime we do not fail if the user has upgraded libannocheck to a newer version than we compiled against.


jimbair (approved):
Assuming this function will always be present, this looks good to me.

dcantrell:
That function will never not be present!  Except when maybe it isn't!

PR #990, opened 2023-01-05
Fix handling of license validation when multiple licensedb files are given

dcantrell:
What was previously happening is that rpminspect would check every license database rather than stop once it had matched every token.


jimbair (approved):
I had to expand a few nearby blocks to follow along, but the change looks good to me.

PR #992, opened 2023-01-06
[lib] When an input cannot be found anywhere, tell the user

dcantrell:
Inputs can be local Koji builds, local RPMs, local SRPMs, remote Koji buids, remote Koji tasks, remote RPMs, remote SRPMs...there's a lot. rpminspect will figure out where the inputs are and copy or download them to the working directory and then begin work.  In cases where an input cannot be located, gather_builds() returns -1 but it doesn't say what happened.  The most common case here is a user specifying a Koji build but Koji returns nothing.  Add a small message indicating the input could not be located.

### DIFF
--- a/lib/builds.c
+++ b/lib/builds.c
@@ -926,6 +926,7 @@ int gather_builds(struct rpminspect *ri, bool fo)
                 return r;
             }
         } else {
+            warnx(_("unable to locate after build: %s"), ri->after);
             free_koji_task(task);
             free_koji_build(build);
             return -1;
@@ -997,6 +998,7 @@ int gather_builds(struct rpminspect *ri, bool fo)
             return r;
         }
     } else {
+        warnx(_("unable to locate after build: %s"), ri->after);
         free_koji_task(task);
         free_koji_build(build);
         return -1;


jimbair (approved):
I 100% approve of this, though the more user-helpful the message the better. A few examples:

1. Unable to locate local build %s
2. Unable to communicate with koji instance
3. Koji instance doesn't have the requested build %s

The idea that if it's missing it says as much, and if talking to a remote system, it differentiates between a successful handshake and expected 404 type error vs koji being down (or dns failing etc).

Just some thoughts but any helpful user feedback is good in my book.

dcantrell:
I agree with the improved output.  The way I wrote the function is that it tries everything and falls through in order of:  local RPM or Koji build, then it tries to ask the Koji server for the build, finally it tries to ask the Koji server for it as a task.  There's not an easy way to distinguish between these types of inputs.

I think what I could do is add something in that indicating whether or not it could talk to the Koji instance.  And if not, display a message above this one.  So if you get both, you can kind of see it was a network communication failure.

dcantrell:
Well, I'm not sure that's going to work either.  Since I don't have a way to determine what the input type is, I can't make the error here more detailed.  I suppose rpminspect could grow options like --before-koji-build= and --after-koji-build=, but that's getting super complicated and would break existing behavior.  Not sure I can do anything else here right now.  But maybe in the future.

PR #993, opened 2023-01-06
[lib] Expand xmlrpc fault reporting for getBuild and getTaskInfo

dcantrell:
Expand the xmlrpc fault reporting here for the initial getBuild and getTaskInfo operations.  This happens from time to time, especially if Koji returns a result where a struct member changes type.  This will hopefully make debugging in the future easier.


jimbair (approved):
Good addition to error reporting

Issue #991, opened 2023-01-06
debuginfo: Missing: .gdb_index


msrb:
**Describe the bug**

debuginfo inspection complains about missing `.gdb_index` section in binaries that are in non-debuginfo RPMs. Note this only happens when the "rhel-8" profile is active.

**To Reproduce**
```
$ rpminspect-redhat -T debuginfo -a x86_64 --profile rhel-8 cups-2.2.6-51.el8
debuginfo:
----------
1) /usr/bin/cupstestdsc in cups-1:2.2.6-51.el8 on x86_64 is missing ELF sections
 

Result: BAD
Waiver Authorization: Anyone
...
```

```
$ rpm -qa| grep rpminspect
rpminspect-data-fedora-1.10-0.1.202210271518git4429e19.fc37.noarch
librpminspect-1.11-0.1.202301051718git2bae93f.fc37.x86_64
rpminspect-1.11-0.1.202301051718git2bae93f.fc37.x86_64
rpminspect-data-redhat-1.8-0.1.202301051724git10cc7d9.fc37.noarch
rpminspect-data-centos-1.3-0.1.202301051740gita051b6d.fc37.noarch

```

**Expected behavior**
I am actually not sure, maybe the current behavior is expected, but then many (majority?) of RHEL 8 builds started failing debuginfo inspection when the rhel-8 profile was tweaked.

**Environment:**
 - RHEL

This was originally reported downstream as OSCI-4299.

msrb:
`rhel-8` profile contains the following section:

```
debuginfo:
    debuginfo_sections: .gdb_index
```

dcantrell:
The source of this distinction in what rpminspect checks for comes from this in rpminspect's ancestor:

```
/* for EL8 and later builds, we need to check for .gdb_index in debuginfo, .gnu.debugdata elsewhere */
```

The way we build things changed starting with el8.  But what I think is happening now is that rpminspect is failing to distinguish between debuginfo and non-debuginfo files.  Continuing to investigate.

dcantrell:
OK, I did some more digging and I think see what I did.  Patch for rpminspect coming as well as a patch for rpminspect-data-redhat for RHEL-7 and prior configurations.

PR #994, opened 2023-01-09
[fix] Correct how the debuginfo checks for missing sections

dcantrell:
The logic in this inspection was half-right or half-wrong, depending on your perspective.  After going back through this and how the predecessor to rpminspect handled it, I realized I was doing the wrong thing on files in regular packages.

I refactored the inspection and updated the test cases and now things should be working as expected.

NOTE: This change does require an update to the vendor data package. Specifically the debuginfo_sections setting in the config file.

Fixes: #991

### DIFF
--- a/data/generic.yaml
+++ b/data/generic.yaml
@@ -1062,6 +1062,6 @@ debuginfo:
 
     # The ELF section name(s) required in debuginfo packages.  The
     # default is shown here.  Some older releases may have to change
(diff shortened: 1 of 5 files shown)


jimbair (approved):
This looks good to me; I don't have anything to add or critique; no nits like typos either. The logic flipping seems to track with your description as well. Nice catch on the fix/refactor.

PR #1000, opened 2023-01-09
[lib] Improve the failure reporting from gather_builds()

dcantrell:
Followup to last week's patch.  This rearranges how the program attempts to figure out what you asked it to download so it can report failures more intelligently.  Or at least appear that way.


jimbair (approved):
I always love extra error reporting =)

PR #1012, opened 2023-01-14
[fix] Modularity support detection

msrb:
Don't treat warnings as errors in the modularity detection code.

There are unused variables (argc/argv) and thus the code fails to compile if -Wall flag is also used. For example Fedora and RHEL use the -Wall flag by default.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -130,7 +130,6 @@ rpmtag_src = '''
     }
     '''
 have_modularitylabel = cc.compiles(rpmtag_src,
-                                   args: [ '-Werror' ],
                                    name: 'RPMTAG_MODULARITYLABEL availability test')
 
 if have_modularitylabel


dcantrell:
I am unable to reproduce this problem on RHEL.  I changed to "warning_level=3" which gives -Wall in addition to -Werror and the RPMTAG_MODULARITYLABEL test worked.  Where are you seeing this?  I don't mind cranking up warning_level to 3 by default.  And I can change the test to int main(void).

msrb:
Sorry, I should have been more specific. `%meson` macro in the spec file sets `-Wall`.

New rpminspect builds for both Fedora and RHEL have the modularity support disabled:
```
Checking if "RPMTAG_MODULARITYLABEL availability test" compiles: NO 
```
This is for example RHEL 9: https://download.copr.fedorainfracloud.org/results/dcantrell/rpminspect/epel-9-x86_64/05224940-rpminspect/build.log.gz

dcantrell:
OK, thanks for the clarification.

PR #1013, opened 2023-01-17
Update secrule description comment

thoger:
Commit dde6edc069 extended the security rule format to include path, increasing the number of fields from four to five.  The mention of the number of fields in the comment of the template file was missed and not updated.


dcantrell:
Thanks!

PR #1015, opened 2023-01-17
[fix] Add back support for older libcurl versions

dcantrell:
At least on CentOS 7 and similar systems, the version of libcurl lacks CURLINFO_CONTENT_LENGTH_DOWNLOAD_T.  So add back support for that with detection in meson.build since they are part of an enum and not a macro.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -67,6 +67,24 @@ clamav = dependency('libclamav', required : true)
 icu_uc = dependency('icu-uc', required : true)
 icu_io = dependency('icu-io', required : true)
 
+# Check for newer CURLcode in libcurl
+curlinfo_src = '''
+    #include <curl/curl.h>
+    int main(void)
+    {
+        CURLcode info = CURLINFO_CONTENT_LENGTH_DOWNLOAD_T;
+        return 0;
+    }
+    '''
+have_newer_curlinfo = cc.compiles(curlinfo_src,
+                                  args: [ '-Werror' ],
+                                  dependencies: [ libcurl ],
+                                  name: 'CURLINFO_CONTENT_LENGTH_DOWNLOAD_T availability test')
+
+if have_newer_curlinfo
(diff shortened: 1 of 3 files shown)


meson.build
```diff
@@ -67,6 +67,24 @@ clamav = dependency('libclamav', required : true)
...
 
+# Check for newer CURLcode in libcurl
+curlinfo_src = '''
+    #include <curl/curl.h>
+    int main(void)
+    {
+        CURLcode info = CURLINFO_CONTENT_LENGTH_DOWNLOAD_T;
+        return 0;
+    }
+    '''
```
eli-schwartz, line 78: Any particular reason not to use has_header_symbol() here? Then you just need the header name and (enum) symbol name.

```
cc.has_header_symbol('curl/curl.h', 'CURLINFO_CONTENT_LENGTH_DOWNLOAD_T', dependencies: [ libcurl ])
```

debug log:
```
Running compile:
Working directory:  /tmp/builddir/meson-private/tmpwv5ej0s6
Command line:  cc /tmp/builddir/meson-private/tmpwv5ej0s6/testfile.c -o /tmp/builddir/meson-private/tmpwv5ej0s6/output.obj -c -D_FILE_OFFSET_BITS=64 -O0 

Code:
 
        #include <curl/curl.h>
        int main(void) {
            /* If it's not defined as a macro, try to use as a symbol */
            #ifndef CURLINFO_CONTENT_LENGTH_DOWNLOAD_T
                CURLINFO_CONTENT_LENGTH_DOWNLOAD_T;
            #endif
            return 0;
        }
Compiler stdout:
 
Compiler stderr:
  
```
console log:
```
Header "curl/curl.h" has symbol "CURLINFO_CONTENT_LENGTH_DOWNLOAD_T" with dependency libcurl: YES  
```

PR #1056, opened 2023-02-01
Fix tests for virus inspection

msrb:
2 changes:

- virus tests are currently not recognized as tests due to the timeout decorator being used on the whole class
  - decorating method instead fixes this problem
- switch to `check_results()` in `TestRPMs` as virus inspection results starts with information about the database and not actual virus scan result

### DIFF
--- a/test/baseclass.py
+++ b/test/baseclass.py
@@ -564,14 +564,9 @@ class TestRPMs(RequiresRpminspect):
                 self.dumpResults()
 
             self.assertEqual(self.p.returncode, self.exitcode)
-            self.assertEqual(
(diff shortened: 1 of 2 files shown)


msrb:
`UnicodeDecodeError: 'utf-8' codec can't decode byte 0xa3 in position 4109: invalid start byte` in tests is because of https://github.com/rpminspect/rpminspect/pull/1057.

msrb:
@dcantrell Please check now ;)

dcantrell:
Conflicts present, I can't merge.

msrb:
Rebased.

PR #1057, opened 2023-02-01
[fix] Nullify result params after they were freed

msrb:
`add_result_entry()` is reading non-NULL params and thus we might end up with random garbage in the final output. Virus inspection tests are currently failing on unicode decoding errors because of this.


dcantrell:
Nice, thank you!

Issue #1059, opened 2023-02-01
rpminspect-centos is not able to compare scratch-builds


mmuzila:
```
$ rpminspect-centos 1835816 1866351
rpminspect: unable to download after task: 1866351
```

```
$ koji -p stream watch-task 1835816
Watching tasks (this may be safely interrupted)...
1835816 build (c9s-candidate, openblas-0.3.21-1.el9.src.rpm): closed
  1835820 buildArch (openblas-0.3.21-1.el9.src.rpm, aarch64): closed
  1835822 buildArch (openblas-0.3.21-1.el9.src.rpm, i686): closed
  1835821 buildArch (openblas-0.3.21-1.el9.src.rpm, ppc64le): closed
  1835824 buildArch (openblas-0.3.21-1.el9.src.rpm, s390x): closed
  1835817 rebuildSRPM (noarch): closed
  1835823 buildArch (openblas-0.3.21-1.el9.src.rpm, x86_64): closed

1835816 build (c9s-candidate, openblas-0.3.21-1.el9.src.rpm) completed successfully
```
```
$ koji -p stream watch-task 1866351
Watching tasks (this may be safely interrupted)...
1866351 build (c9s-candidate, openblas-0.3.21-2.el9.src.rpm): closed
  1866381 buildArch (openblas-0.3.21-2.el9.src.rpm, aarch64): closed
  1866383 buildArch (openblas-0.3.21-2.el9.src.rpm, i686): closed
  1866382 buildArch (openblas-0.3.21-2.el9.src.rpm, ppc64le): closed
  1866385 buildArch (openblas-0.3.21-2.el9.src.rpm, s390x): closed
  1866352 rebuildSRPM (noarch): closed
  1866384 buildArch (openblas-0.3.21-2.el9.src.rpm, x86_64): closed

1866351 build (c9s-candidate, openblas-0.3.21-2.el9.src.rpm) completed successfully
```

dcantrell:
I've tracked this down to the curl_get_size() that I have.  It's not following HTTP redirects which the CentOS Koji infrastructure is making use of for downloads.  I need to patch that and get it working.  Right now it's getting -1 back for every size which is also why you're seeing the error about needing 16 EiB of available space.

Working on this today.

PR #1063, opened 2023-02-01
[fix] Set CURLOPT_FOLLOWLOCATION in curl_get_size()

dcantrell:
For redirects, make sure curl follows along in order to get the size of the file we're trying to download.

Fixes: #1059

### DIFF
--- a/lib/curl.c
+++ b/lib/curl.c
@@ -244,11 +244,13 @@ curl_off_t curl_get_size(const char *src)
 
     /* get the size */
     curl_easy_setopt(c, CURLOPT_URL, src);
-    curl_easy_setopt(c, CURLOPT_NOBODY, 1);
+    curl_easy_setopt(c, CURLOPT_NOBODY, 1L);
     curl_easy_setopt(c, CURLOPT_FAILONERROR, true);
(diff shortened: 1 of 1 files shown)


lib/curl.c
```diff
@@ -244,11 +244,13 @@ curl_off_t curl_get_size(const char *src)
 
     /* get the size */
     curl_easy_setopt(c, CURLOPT_URL, src);
-    curl_easy_setopt(c, CURLOPT_NOBODY, 1);
+    curl_easy_setopt(c, CURLOPT_NOBODY, 1L);
     curl_easy_setopt(c, CURLOPT_FAILONERROR, true);
+    curl_easy_setopt(c, CURLOPT_FOLLOWLOCATION, 1L);
```
evan-goode, line 249: Consider setting a maximum number of redirects too, the default is unlimited:

```
curl_easy_setopt(c, CURLOPT_MAXREDIRS, 10L);
```

Was always a bit annoyed FOLLOWLOCATION is not the default, I've had a few things break after something is moved to a new URL and my script doesn't have `curl -L`.

dcantrell: Oh nice, I didn't know that.  Yeah, definitely something worth setting.

PR #1080, opened 2023-02-11
Fix handling with multiple license databases defined

dcantrell:
This required a lot of work, which is why there are a lot of commits.  The business end of the PR is in inspect_license.c.  This preserves functionality of all existing test cases while making sure multiple license db files defined in the config file work as expected.

### DIFF
--- a/lib/inspect_license.c
+++ b/lib/inspect_license.c
@@ -66,7 +66,9 @@ static bool lic_cb(const char *license_name, void *cb_data)
     const char *lic = data->lic;
     parser_plugin *p = data->p;
     parser_context *db = data->db;
-    char *fedora_abbrev = NULL, *fedora_name = NULL, *spdx_abbrev = NULL;
+    char *fedora_abbrev = NULL;
+    char *fedora_name = NULL;
+    char *spdx_abbrev = NULL;
     char *approved = NULL;
 
     if (data->valid) {
@@ -95,8 +97,7 @@ static bool lic_cb(const char *license_name, void *cb_data)
     }
 
     /* No full tag match and no abbreviations; license entry invalid. */
-    if (strlen(fedora_abbrev) == 0 && strlen(spdx_abbrev) == 0 &&
-        strlen(fedora_name) == 0) {
+    if (strlen(fedora_abbrev) == 0 && strlen(spdx_abbrev) == 0 && strlen(fedora_name) == 0) {
         goto done;
     }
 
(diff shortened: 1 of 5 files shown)


jimbair:
A lot of logic changes, though they seemed easy enough to follow. Comments and flow all seem to make sense to me; no objections or questions from my side.Though it seems like a few checks are having a bad time.

jimbair (approved):

lib/parse_json.c
```diff
@@ -82,6 +82,7 @@ static json_object *getobj(json_object *jo, const char *key1, const char *key2)
         assert(key2 == NULL);
         return jo;
     } else if (!json_object_is_type(jo, json_type_object)) {
+        warnx("json_object_is_type is %s for %s", json_type_to_name(json_object_get_type(jo)), key1);
```
frozencemetery, line 85: So unfortunately, this causes test_cfg to fail because it's comparing parser output, and this branch gets hit.

This is due to a design choice that json-c made around NULL objects.  For a `json_object *jo`, the fact that `jo == NULL` means that it's unitialized.  However, if `json_object_get_type(jo) == json_type_null`, that means that jo is equivalent to "not found in file", or was specified as literal null.  (These are basically equivalent.)

Adding to this confusion: the type of a `json_object` may or may not be `json_type_object`.  Our parser interface hides this weirdness and instead only talks about "dicts" and "arrays" in the hope of not being ambiguous.

In the YAML parser, there's a rudimentary trace system - one can `#define DEBUG 1` and get more information about what's happening.  This could be expanded to the other parsers if desired: I didn't do that only because it wasn't needed to make them work.  If getting this output is required without rebuilding rpminspect with different flags, it might also work to use an environment variable to turn on debug printing - like $KRB5_TRACE does.

src/rpminspect.c
```diff
@@ -267,18 +261,20 @@ static char *get_product_release(string_map_t *products, const favor_release_t f
     free(before_product);
 
     /* trim the leading period */
-    i = 0;
+    if (after_product) {
+        i = 0;
```
frozencemetery, line 265: Preexisting, but `i` was already 0 here, regardless of the value of `after_product`.

lib/inspect_license.c
```diff
@@ -139,52 +139,39 @@ static bool check_license_abbrev(parser_plugin *p, parser_context *db, const cha
...
-
-    /* see if the entire substring matches */
-    if (check_license_abbrev(p, db, s)) {
-        return true;
-    }
+    assert(license != NULL);
 
-    /* fall back on checking each individual token */
-    tagtokens = tagcopy = strdup(s);
+    tagtokens = tagcopy = strdup(license);
```
frozencemetery, line 172: Nit: the return from `strdup` is checked above (line 150) but not here.

frozencemetery (changes requested):
Generally looks good, but see inline for comment about the test failure.

lib/parse_json.c
frozencemetery: Since the json parser currently has the luxury of not needing to support backward compat, another possibility is to modify this to only print the warning when the type also isn't null.  That would only catch type errors.

src/rpminspect.c
dcantrell: Thanks

PR #1106, opened 2023-02-28
[ci] Add 'debug' and 'setup-debug' targets to Makefile

dcantrell:
These are for the CI jobs.  For the test suite, I want the CI jobs to run "make check" rather than using meson.  The reason is the Makefile target sets up the test environment and that's what I use to test things.

I have converted most workflows over to running "make debug" and "make check" in the GitHub workflow file, but I still have CentOS and FreeBSD to update.  Need to make sure these work first.

### DIFF
--- a/.github/workflows/almalinux.yml
+++ b/.github/workflows/almalinux.yml
@@ -62,7 +62,6 @@ jobs:
                   make instreqs
 
                   # Build the software and run the test suite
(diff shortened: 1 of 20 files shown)


jimbair (approved):
I assume there's a reason we didn't have meson test / now don't have make check for aarch64/ppc64le/s390x, but they match, so that seems good to me.

PR #1109, opened 2023-02-28
Assortment of CI job fixes

dcantrell:
More things as I clean up the failing jobs.  A lot of this is just Makefile restructuring and job refactoring in the GitHub syntax.

### DIFF
--- a/.github/workflows/freebsd.yml
+++ b/.github/workflows/freebsd.yml
@@ -76,7 +76,7 @@ jobs:
             - name: Check exit codes
               run: |
                   for ec in exitcodes/* ; do
-                      code="$(cat ${ec} 2>/dev/null)"
+                      code="$(cat "${ec}" 2>/dev/null)"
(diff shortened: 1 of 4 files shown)


jimbair (approved):

.github/workflows/freebsd.yml
```diff
@@ -75,7 +75,8 @@ jobs:
             # Process the exit codes outside the VM
             - name: Check exit codes
               run: |
-                  for ec in build/exitcodes/* ; do
-                      code=$(cat ${ec} 2>/dev/null)
+                  for ec in exitcodes/* ; do
+                      code="$(cat ${ec} 2>/dev/null)"
```
evan-goode, line 79: Won't matter, but `${ec}` should be quoted
```
code="$(cat "${ec}" 2>/dev/null)"
```

evan-goode (changes requested):

PR #1111, opened 2023-03-01
Handle 32-bit x86 architecture naming inconsistencies and patch file names with RPM macros

dcantrell:
A lot in here, but it all started with fixing the annocheck inspection and it not finding the debuginfo path for i686 packages.  That began the yak shaving.

### DIFF
--- a/lib/files.c
+++ b/lib/files.c
@@ -159,7 +159,7 @@ rpmfile_t *extract_rpm(struct rpminspect *ri, const char *pkg, Header hdr, const
 {
     rpmtd td = NULL;
 
-    const char *rpm_path;
+    const char *rpm_path = NULL;
     struct file_data *path_table = NULL;
     struct file_data *path_entry = NULL;
     struct file_data *tmp_entry = NULL;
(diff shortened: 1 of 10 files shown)


jimbair (approved):

lib/files.c
```diff
@@ -305,7 +307,27 @@ rpmfile_t *extract_rpm(struct rpminspect *ri, const char *pkg, Header hdr, const
         }
 
         /* Prepend output_dir to the path name */
-        xasprintf(&file_entry->fullpath, "%s/%s", *output_dir, archive_path);
+        tmp = archive_path;
```
evan-goode, line 310: Will `joinpath` not do this?

evan-goode (approved):

lib/files.c
dcantrell: It should.  My initial run it didn't work, so I didn't think much of it.  joinpath.c comes from my secret project and I didn't want to dig in to it because I was already busy shaving this side of the yak.

dcantrell: Oh right, I know why.  archive_path is from a libarchive structure and joinpath() wants allocated strings it can modify and walk independently.  It would work if I strdup()'ed them, but I just decided to do it this way instead.

dcantrell: The stack, and all.

Issue #1112, opened 2023-03-02
[error message improvement] lack of clamav database is not communicated clearly


t184256:
**Describe the bug**
If one tries `-TALL` but doesn't have clamav (clamav-data?) installed,
they're gonna get a rather cryptic `rpminspect: opendir: No such file or directory`.
I know it's a bit of a silly request,
but an error message that mentions clamav would've saved me
good 15 minutes of my time.

**To Reproduce**
```
dnf remove clamav-data
rpminspect -T ALL -k -c /usr/share/rpminspect/fedora.yaml crypto-policies-20221215-1.gita4c31a3.fc38
```
strace reveals that this is preceded with
```
openat(AT_FDCWD, "/var/lib/clamav", O_RDONLY|O_NONBLOCK|O_CLOEXEC|O_DIRECTORY) = -1 ENOENT (No such file or directory)
```

**Expected behavior**
Error message mentions that it couldn't find clamav.

**Environment:**
 - rpminspect-1.11-0.1.202303021442gitf47b071.fc37.x86_64

dcantrell:
Patch coming that changes that to:

```
rpminspect: missing /var/lib/clamav: No such file or directory
```

PR #1113, opened 2023-03-02
[lib] Improve the error reporting when missing /var/lib/clamav

dcantrell:
Or whatever the clamav database directory is.  If the system lacks that directory, report it in a meaningful way for the user.

Fixes: #1112


jimbair (approved):
The opendir text being replaced with missing %s seems like a helpful addition when things go sideways. +1

PR #1123, opened 2023-03-16
[fix] Handle unchanged static_context in modularity test correctly

### LINKED ISSUE #1118: A modularity static_context test fails for RHEL-8 modules although both modules have no static_context
A gating test run as "--release=el8.6 --profile=rhel-8 mysql-8.0-820190104140943.9edba152 mysql-8.0-8080020230313165100.63b34585" <http://artifacts.osci.redhat.com/testing-farm/295d0f3f-656b-4c61-bd3e-4f2ff00876ac/> reports a VERIFY failure on static_context modularity test:

~~~~
The /data/static_context value in mysql-8.0-820190104140943.9edba152 matches the value in mysql-8.0-8080020230313165100.63b34585, but the product release rules forbid the presence of /data/static_context in the module metadata.
~~~~

Both modules indeed have static_context equal, the field is missing, hence defaults to false. And the policy indeed forbids presence of /data/static_context.  Thus the resolution should be a pass. Not a fail.

It looks like a logic bug in rpminspect code. It does not distinguish asc/bsc false from true. See check_static_context():

~~~~
    /* get the after build static_context first */
    asc = get_static_context(ri->worksubdir, AFTER_SUBDIR);

    if (ri->before) {
        /* comparing builds, verify after build is correct, but report changes */
        bsc = get_static_context(ri->worksubdir, BEFORE_SUBDIR);

        if (bsc == asc) {
            if (ri->modularity_static_context == STATIC_CONTEXT_FORBIDDEN) {
                xasprintf(&params.msg, _("The /data/static_context value in %s matches the value in %s, but the product release rules forbid the presence of /data/static_context in the module metadata."), ri->before, ri->after);
                params.severity = RESULT_VERIFY;
                params.waiverauth = WAIVABLE_BY_ANYONE;
                params.remedy = REMEDY_MODULARITY_STATIC_CONTEXT;
                r = false;
            } else if (ri->modularity_static_context == STATIC_CONTEXT_REQUIRED) 
~~~~

This check was added per issue #803.

ppisar:
Previously, if both old and new module build had the same static_context value, the test ignored the value and reported a pass or a failure regadrless of the value.

This patch splits a check for the static_context change from a check for the desired value of the "after" module build. That makes the logic and the code simpler at the expense of gluing two sentences together.

Since the two senteces are independent, I don't think it's a problem for translators. It's similar to separating messages by a new line. But rpminspect uses single-line reports.

Fixes #1118

### DIFF
--- a/lib/inspect_modularity.c
+++ b/lib/inspect_modularity.c
@@ -102,6 +102,7 @@ static bool check_static_context(struct rpminspect *ri)
     bool bsc = false;
     bool asc = false;
     struct result_params params;
+    char *after_compliance;
 
     assert(ri != NULL);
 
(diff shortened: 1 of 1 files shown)


ppisar:
I have two comments to the code which I did not addressed in this pull request to keep it minimal:

There are now test for modularity checks in ./test directory.
And the wording of this static_context check is misleading. The check does test whether /data/static_context is present. It tests (and should test) whether /data/static_context is (true), or (false or missing).

dcantrell:
Well you were working on this today and I was too.  My patch is way way smaller than yours.  Which one is correct?  :)

ppisar:
I think your new code does not handle a case when both modules have static_context: false (!bsc && !asc). In that case the policy is not consulted, no message is reported to the user, and the test always passes. So it does not fix this issue at all. It only inverts the resolution from always fail to always pass.

A design flaw of your code is that the for policy cases in (!ri->before) branch have to be repeated for each bsc/asc combination in (ri->before) branch. That means 4 times 4.

My patch is larger, because it shuffles more lines, but number of the resulting lines is smaller. It only requires 5 + 3 branches. And what's more important, the policy checking is not copied over multiple places.

ppisar:
I'll try not to forget to leave a notice next time.

Issue #1026, opened 2023-01-19
Support for security rules for the virus inspection


thoger:
**Describe the bug**
Issue #1025 requests making virus inspection failures WAIVABLE_BY_SECURITY.  If that is done, there's also an associated need to have centralized way to allow-list certain acceptable failures (e.g. when eicar test file is included in a test suite of some package).  Adding a "virus" rule type to the existing security rules mechanisms would make sense.

dcantrell:
I'm thinking in the data/security/ files, adding "virus=" which can be SKIP, INFORM, VERIFY, or FAIL.  I believe this is what you described, but I just want to confirm.  If this is sufficient, the modification will be easy to rpminspect.

Thoughts @thoger ?

PR #1136, opened 2023-04-04
Implement the 'virus' security rule type

dcantrell:
As described in #1026, this implements a 'virus' security rule type that can be SKIP, INFORM, VERIFY, or FAIL per file or path glob pattern.  This works for source and binary packages.


evan-goode (approved):

snbueno (approved):
All looks good to me!

PR #1142, opened 2023-04-10
[tst] Skip virus tests on FreeBSD

dcantrell:
The FreeBSD CI job on GitHub Actions is a VM in a macos guest which I think is a container in a VM itself.  The virus inspection nearly always times out, so just skip it on this platform.

### DIFF
--- a/.github/workflows/freebsd.yml
+++ b/.github/workflows/freebsd.yml
@@ -64,11 +64,15 @@ jobs:
                       gmake instreqs
                       echo $? > exitcodes/instreqs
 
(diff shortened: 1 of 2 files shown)


evan-goode:
Maybe a nit, but it would be nice to disable these tests in the GitHub action, rather than in the tests themselves, so someone could still run them on their FreeBSD system.

dcantrell:
That's not unreasonable.

evan-goode (approved):

jimbair (approved):

PR #1144, opened 2023-04-12
[fix] Include libgen.h in init.c (for basename)

dcantrell:
Implicit definition is forbidden by C99.


jimbair (approved):
I suppose I'll let this one slide by.

PR #1148, opened 2023-04-18
[lib] Minor porting work to get the software building on NetBSD

dcantrell:
This helped clean up some of the ELF checking code as well as more about meson.build files and their limitations and, I guess, features.


jimbair (approved):

lib/inspect_elf.c
```diff
@@ -18,13 +18,17 @@
...
-
-#include <gelf.h>
 #include <libelf.h>
 
+#ifndef PT_GNU_STACK
+#define PT_GNU_STACK 0x647e551 /* Indicate executable stack */
+#endif
+
+#ifndef PT_GNU_RELRO
+#define PT_GNU_RELRO 0x647e552 /* Make read-only after relocation */
```
snbueno, line 29: More for my own edification than anything, but where does this (and the above) memory address come from?

snbueno (approved):
I think these changes look good, as best as I can see at least. Nice list of supported OSes you're building. :)

evan-goode (approved):

lib/inspect_elf.c
dcantrell: PT_GNU_RELRO is one of the legal values for p_type in an Elf64_Phdr.  On a glibc Linux system the values are defined in /usr/include/elf.h.  Non-Linux systems vary with regards to what they define.  On NetBSD they have defined p_type values from GNU systems in /usr/include/elf.h, but I can't include elf.h on NetBSD because it collides with type definitions from libelf installed from pkgsrc.  libelf is similar to our libelf in elfutils, but the headers to not coexist with the NetBSD system headers.  Because I only needed two p_type values in inspect_elf.c, I decided to define them if they did not get pulled in from a header.  It actually gets worse because libelf that NetBSD has in pkgsrc lacks gelf_getsymshndx() which means I need to figure out a way to do the same thing but with only gelf_getsym().  I didn't do that in this PR.

dcantrell:
> I think these changes look good, as best as I can see at least. Nice list of supported OSes you're building. :)

It's surprising how these other platforms are turning up minor things here and there and that leads to finding problems in other libraries on Linux.  Certainly the gcc vs. clang stuff.

However, if you thought NetBSD would be at all similar to FreeBSD, you would be mistaken.

lib/inspect_elf.c
dcantrell: Also, here's everything you didn't want to know about ELF:

https://en.wikipedia.org/wiki/Executable_and_Linkable_Format

A whole lot of what we do checking built packages is examining what ends up in an ELF file.  I don't know a lot about the other formats still in use today (PE on Windows and Mach-O on MacOS), but I also don't care since we are dealing with ELF only.

snbueno: Ahhhaha, ok. I figured you had to have a good reason for dropping elf.h from your list of includes (and so didn't question it), and this explains it. Makes much more sense now.

Issue #1155, opened 2023-04-20
RFE: virus inspection: show also version of daily.cvd


msrb:
**Is your feature request related to a problem? Please describe.**

rpminspect virus inspection currently shows the following information about clamav and its virus database:

```
clamav version 0.103.8
/var/lib/clamav/main.cvd version 62 (16 Sep 2021 08-32 -0400)
/var/lib/clamav/bytecode.cvd version 334 (22 Feb 2023 16-33 -0500)
```

However, there is also `daily.cvd`, which is updated daily and it contains signatures for the latest threats. It would be good to have the version of `daily.cvd` also caputured in the inspection output.
https://blog.clamav.net/2021/03/clamav-cvds-cdiffs-and-magic-behind.html

**Describe the solution you'd like**

```
clamav version 0.103.8
/var/lib/clamav/main.cvd version 62 (16 Sep 2021 08-32 -0400)
/var/lib/clamav/daily.cvd version 26825 (20 Apr 2023 09-21 -0500)
/var/lib/clamav/bytecode.cvd version 334 (22 Feb 2023 16-33 -0500)
```

PR #1158, opened 2023-04-24
[lib] Load *.cld files in addition to *.cvd files for virus scans

dcantrell:
clamav databases are either *.cvd files or *.cld files so make sure to load both if we find them.

Fixes: #1155


jimbair (approved):

msrb:
Thanks! ;)

PR #1160, opened 2023-04-26
[fix] Load both *.cld and *.cvd virus database files, if they exist

msrb:
This just fixes a logical error that was accidentally introduced in 66b0953.


dcantrell:
Whoops.  :)

Well this week my brain isn't exactly working as it usually does.

msrb:
> Well this week my brain isn't exactly working as it usually does.

I share that feeling :sweat_smile:

Issue #1157, opened 2023-04-24
Excluding files for unicode inspection needs to take into account where the source archive is extracted


jchecahi:
**Describe the bug**
Unicode inspection unpacks some source archives manually (not using rpm API) in a separate directory such as `/var/tmp/rpminspect/local.XXXXXX/rpmbuild/BUILD/unpack-XXXXXX`.  Looking to #1073 it seems like it shouldn't be taking that last `unpack-XXXXXX` directory into account for reporting. Also not clear to me, but looks like it shouldn't take that into account for ignored files.

I'm trying to add an unicode exception for a file inside one of those manually unpacked source files (clang-tools-extra-16.0.1.src), and I must specify `unpack-*/clang-tools-extra-*.src/<rest_of_path>` in order to make it work.

Looking into the source, it seems like [this check](https://github.com/rpminspect/rpminspect/blob/959e480c414b01468dd0529f3a3542437559208d/lib/inspect_unicode.c#L429) assumes that `build` contains the whole path up to `unpack-XXXXXX` but instead point to its parent dir: `BUILD`, but I might be wrong.

Hence adding this to `security/el8` in `rpminspect-data-<vendor>` (in my case `rpminspect-data-redhat`) makes the inspection to *not* ignore the file:
```
clang-tools-extra-*.src/docs/clang-tidy/checks/misc/misleading-bidirectional.rst  clang  *  *  unicode=INFORM
``` 
While line makes the inspection to ignore the file and hence pass the unicode inspection:
```
unpack-*/clang-tools-extra-*.src/docs/clang-tidy/checks/misc/misleading-bidirectional.rst  clang  *  *  unicode=INFORM
```

**To Reproduce**
```
rpminspect-redhat -v -w ~/builds -f rhel8-8080020230119204057.fd72936b
rpminspect-redhat -k -v -T unicode -a src ~/builds/llvm-toolset-rhel8-8080020230119204057.fd72936b
```

**Expected behavior**
It should be possible to add ignored files without taking into account whether the source archive will be extracted manually in a `rpmbuild/BUILD/unpack-XXXXXX` directory or directly into `rpmbuild/BUILD`.

dcantrell:
This reads like a bug.  My intent is to make source tree specification relative to an invisible root tree but relative to the extracted SRPM.  Within that you would have extracted and patched source archives.  So think of it as a tree that would look like the result of `fedpkg prep` in a dist-git repo in Fedora.

[...]

PR #1161, opened 2023-04-27
[fix] Handle manually unpacked source archives in unicode inspection

dcantrell:
The unicode inspection works on SRPMs and needs to both look at the files in the SRPM itself as well as unpacked and patched source trees. Archives are unpacked to temporary directories and the code was not properly trimming this leading path so that matching against security/PRODUCT_RELEASE rules in the vendor data package would work.

Fixes: #1157

### DIFF
--- a/lib/inspect_unicode.c
+++ b/lib/inspect_unicode.c
@@ -219,6 +219,16 @@ static char *rpm_prep_source(struct rpminspect *ri, const rpmfile_entry_t *file,
             warn("close");
         }
 
+        if (close(STDOUT_FILENO)) {
+            warn("close");
+            _exit(EXIT_FAILURE);
+        }
+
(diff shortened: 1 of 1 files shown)


jimbair (approved):
I see this also includes the removal of ignoring certain paths (I assume defined by the user or config); either way, ignoring files/paths in this inspection seems like a bad idea I think so no complaints really.

dcantrell:
> I see this also includes the removal of ignoring certain paths (I assume defined by the user or config); either way, ignoring files/paths in this inspection seems like a bad idea I think so no complaints really.

It does.  Those calls ended up being a no-op but I feel removing them is a better idea.  The only way to affect the unicode inspection rules is modifying the vendor data package.

PR #1165, opened 2023-05-04
[tst] Add unit test for arches.c

zmiklank:
The unit test in this PR in the end covers only `allowed_arch()`.

The allowed arches unit test adds one allowed arch and then checks if the added arch is indeed among allowed arches. It also checks, that a random not-allowed arch is indeed not allowed.


dcantrell:
Looks good, thanks!

Issue #1167, opened 2023-05-04
unicode inspection still not stripping `unpack-XXXXXX` directories.


jchecahi:
**Describe the bug**
This is a followup for #1157 . After some testing, unicode inspection is still using the manual prep directory and prints `A forbidden code point, <codepoint>, was found in the unpack-XXXXXX/<source tree> source file....`

**To Reproduce**
```
rpminspect-redhat -v -w ~/builds -f rhel8-8080020230119204057.fd72936b
rpminspect-redhat -k -v -T unicode -a src ~/builds/llvm-toolset-rhel8-8080020230119204057.fd72936b
```

**Expected behavior**
The `unpack-` directories should be transparent to the user and not shown in the report or required in the security/<release> files.

PR #1168, opened 2023-05-04
[fix] unicode inspection: Handle unpack- directories from manual extract

jchecahi:
This commit fixes the localpath so it can later match the UNPACK_BASE to strip the unpack-XXXXXX directory. Also the localpath was not trimming the directory completely as it was using UNPACK_BASE instead UNPACK_TEMPLATE.

Fixes: #1167


lib/inspect_unicode.c
```diff
@@ -455,13 +455,17 @@ static int validate_file(const char *fpath, __attribute__((unused)) const struct
          *     rpminspect-1.47.0/lib/magic.c
          */
         localpath += strlen(build);
```
dcantrell, line 457: Just one style comment here, but can you put a blank line between 'localpath += strlen(build);' and the block you added?

dcantrell (changes requested):
This looks good, only one style change requested.  Thanks.

Issue #1166, opened 2023-05-04
%patch -P1 -p1 not recognize


jankaluza:
**Describe the bug**
It seems that `%patch -P1 -p1` is not recognized.

The following `Patch1` is applied here https://gitlab.com/redhat/centos-stream/rpms/rhel-system-roles-sap/-/blob/c9s/rhel-system-roles-sap.spec#L147, but the rpminspect complains about it:

```
Patch number 1 (community.sap_install-shebang.patch) is missing a corresponding %patch1 macro, usually in %prep.
```

**To Reproduce**
$ rpminspect -c /usr/share/rpminspect/redhat.yaml rhel-system-roles-sap-3.5.0-2.el9.src.rpm

https://kojihub.stream.rdu2.redhat.com/koji/buildinfo?buildID=32525

**Expected behavior**
%patch -P1 is supported. Ideally support all the formats mentioned in https://rpm-software-management.github.io/rpm/manual/spec.html.

**Environment:**
 - Fedora 36.

PR #1170, opened 2023-05-04
[fix] Honor additional %patch macro syntax in the patches inspection

dcantrell:
The %patchN macro has been deprecated by rpm, so many maintainers are switching to alternate syntax.  The preferred syntax is:

    %patch -P N

Where N corresponds to the patch number as defined in the preamble. However, other syntax is supported.  rpminspect can now work with the following syntax:

    %patchN
    %patch -P N
    %patch -PN
    %patch N

Fixes: #1166

### DIFF
--- a/include/constants.h
+++ b/include/constants.h
@@ -658,10 +658,17 @@
 /**
  * @def SPEC_MACRO_PATCH
  *
- * The %patchN macro used to apply patches.  N corresponds to the
- * patch number defined in the header.
+ * The %patch macro used to apply patches.
  */
 #define SPEC_MACRO_PATCH       "%patch"
+
+/**
+ * @def SPEC_MACRO_PATCH_P_ARG
+ *
+ * The -P argument to the %patch macro.  The -P argument is used to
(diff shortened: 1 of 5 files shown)


evan-goode:
Looks good, would be nice to also add/modify a couple tests in (I think) https://github.com/rpminspect/rpminspect/blob/main/test/test_patches.py.

Would a warning for the deprecated syntax be appropriate, or I guess that is more of a job for rpmlint?

jimbair (approved):

dcantrell:
> Looks good, would be nice to also add/modify a couple tests in (I think) https://github.com/rpminspect/rpminspect/blob/main/test/test_patches.py.

True.  rpmfluff currently uses only the %patchN syntax, so tests would need to work around that.  But I can construct manual tests for those cases.

> Would a warning for the deprecated syntax be appropriate, or I guess that is more of a job for rpmlint?

Not rpmlint but rpmbuild, which is already emits a warning about the deprecated syntax.

evan-goode (approved):

PR #1171, opened 2023-05-05
[fix] Fix a memory bug in the unicode inspection

msrb:
Double the line buffer, if needed.

rpminspect started freezing in RHEL CI with the following error in logs:
"corrupted size vs. prev_size"

Doing just "sz += (i - sz);" is problematic as we would initially not increase the size of the buffer at all ("sz += 0").

### DIFF
--- a/lib/inspect_unicode.c
+++ b/lib/inspect_unicode.c
@@ -499,7 +499,7 @@ static int validate_file(const char *fpath, __attribute__((unused)) const struct
         for (i = 0; (line[i] = u_fgetcx(src)) != U_EOF && !end_of_line(line[i]); i++) {
             /* increase the buffer size if necessary */
             if (i >= sz) {
-                sz += (i - sz);
+                sz *= 2;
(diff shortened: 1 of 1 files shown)


msrb:
What's funny is that we did not see (afaik) the problem in Fedora. Maybe the initial buffer size (`BUFSIZ`) is larger there, or we were just lucky... :)

dcantrell:
> What's funny is that we did not see (afaik) the problem in Fedora. Maybe the initial buffer size (`BUFSIZ`) is larger there, or we were just lucky... :)

And this is why we test both Fedora and RHEL.  :)

msrb:
It turned out that we hit the problem in Fedora as well:
https://artifacts.dev.testing-farm.io/f3e120fd-f366-424d-90ac-9469cf67bb9d/work-rpminspects1lgwk9y/rpminspect/execute/data/guest/default-0/rpminspect-1/data/verbose.log

PR #1172, opened 2023-05-15
Introduce 'udevrules' inspection

ldv-alt:
The udevrules inspection checks udev rules files syntax using "udevadm verify".  The latter is a new udevadm tool command introduced in systemd v254.

### DIFF
--- a/include/results.h
+++ b/include/results.h
@@ -773,6 +773,15 @@
 
 /** @} */
 
+/**
+ * @def REMEDY_UDEVRULES
+ *
+ * How to address syntax check issues in udev rules files.
+ */
+#define REMEDY_UDEVRULES _("Refer to the udev documentation at https://www.freedesktop.org/software/systemd/man/udev.html for help correcting the errors and warnings.")
+
+/** @} */
+
 /** @} */
 
 #endif
--- a/README.md
+++ b/README.md
@@ -153,6 +153,7 @@ there are a number of userspace programs used:
     /usr/bin/msgunfmt
     /usr/bin/abidiff [optional]
     /usr/bin/kmidiff [optional]
+    /usr/bin/udevadm [optional]
 
 The provided spec file template uses the Fedora locations for these
 files, but in the program, they must be on the runtime system.
--- a/data/generic.json
+++ b/data/generic.json
@@ -252,5 +252,11 @@
       "/usr/lib/grub/*"
     ],
     "debuginfo_sections": ".symtab .debug_info"
+  },
+  "udevrules": {
+    "udev_rules_dirs": [
+      "/etc/udev/rules.d/",
+      "/usr/lib/udev/rules.d/"
+    ]
   }
 }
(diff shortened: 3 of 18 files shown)


include/results.h
```diff
@@ -773,6 +773,15 @@
 
 /** @} */
 
+/**
+ * @def REMEDY_UDEVRULES
+ *
+ * How to address syntax check issues in udev rules files.
+ */
+#define REMEDY_UDEVRULES _("Refer to the udev documentation at https://www.freedesktop.org/software/systemd/man/udev.html for help correcting the errors and warnings")
```
dcantrell, line 781: Need a period after warnings.

dcantrell (changes requested):
Just a string cleanup.

dcantrell:
This looks really nice, thanks for the PR!

It might not be possible, but is 'udevadm verify' available via a library at all or is just only available as the command?

Given the requirement of systemd v254, the spec file template needs to carry a Requires for that.  If the packaging varies across Fedora, CentOS Stream, and RHEL, it might be best to just put the Requires on the individual command at >= 254.

Thanks!

ldv-alt:
> It might not be possible, but is 'udevadm verify' available via a library at all or is just only available as the command?

Unfortunately, systemd doesn't provide a library for parsing udev rules files.

> Given the requirement of systemd v254, the spec file template needs to carry a Requires for that. If the packaging varies across Fedora, CentOS Stream, and RHEL, it might be best to just put the Requires on the individual command at >= 254.

Given that systemd 254 has not been released yet, would it be OK to specify a requirement to an unreleased version?
Something along these lines?
```
# The udevrules inspection requires an external executable (udevadm verify)
# provided by systemd-udev.  If the installed udevadm executable does not
# provide 'verify' command, you can disable the udevrules inspection.
%if 0%{?rhel} >= 8 || 0%{?epel} >= 8 || 0%{?fedora}
Recommends:     systemd-udev >= 254
%else
Requires:       systemd-udev >= 254
%endif
```

ldv-alt:
OK, I've added a runtime check.

ldv-alt:
In the latest edition I've introduced run_udevadm_verify() wrapper to simplify `udevadm verify` invocation.
Also, I added --no-summary option to `udevadm verify` to turn off the summary which just clutters the output.

ldv-alt:
@dcantrell, is anything else expected from me in this PR?

ldv-alt:
@dcantrell, please let me know if something has to be changed in this PR.

dcantrell:
Everything looks good at this point.  I just want to see the CI jobs pass before merging it.

ldv-alt:
> Everything looks good at this point. I just want to see the CI jobs pass before merging it.

Hmm, CI seems to be broken for unrelated reasons.

dcantrell (approved):

dcantrell:
It's the Fedora rawhide Docker image that I use in GHA.  It is missing dnf5 as well as /etc/dnf/dnf.conf for dnf 4.x.  I've fixed that in my CI job definition, but there's also a bug to fix the Docker images for rawhide.

PR #1182, opened 2023-05-18
[fix] Use RPMTAG_NOT_FOUND macro instead of -1

dmnks:
In RPM 4.19, the internal representation of the rpmTagVal type has been changed from signed to unsigned.  That means, a plain check for the value of -1 (tag not found) returned from rpmTagGetValue() now produces a compiler warning.

In order to fix that, we have to either explicitly cast the literal -1 to (uint32_t)-1 or just use the RPMTAG_NOT_FOUND macro readily provided by RPM.

Choose the latter since that's the supported way, really, it's also now documented as such in the API docs for the rpmTagGetValue() function.

### DIFF
--- a/lib/inspect_modularity.c
+++ b/lib/inspect_modularity.c
@@ -181,7 +181,7 @@ static bool modularity_driver(struct rpminspect *ri, rpmfile_entry_t *file)
 
     /* Find how to find the header */
     tv = rpmTagGetValue("modularitylabel");
-    if (tv == -1) {
+    if (tv == RPMTAG_NOT_FOUND) {
         add_result(ri, &params);
         free(params.msg);
         return false;


dmnks:
Just  for some more context, this issue was found while rebuilding rpminspect against RPM 4.19 in Fedora Rawhide: https://kojipkgs.fedoraproject.org//work/tasks/1947/101251947/build.log

dcantrell:
This is nice, but I still need to support older releases as well.  rpminspect builds and runs on platforms back to RHEL-7, which is necessary for internal use for some projects.  I don't mind an ifdef block to deal with the older API vs. the new API.  I can do that or if you want to update the PR, that's fine too.

dmnks:
Understood.

Looking at RPM's git log, the `RPMTAG_NOT_FOUND` define was [added](https://github.com/rpm-software-management/rpm/commit/7e56c6355bea552d89e8d5d0a317763a8cee4a4f) back in 2007, so we should be safe in terms of RHEL-7. There should be no need for ifdefs either.

PR #1196, opened 2023-05-31
[ci] Account for dnf5 presence in Fedora rawhide

dcantrell:
Our GitHub Action for Fedora rawhide builds needs to account for dnf5 presence.  Use that if we find it.  Currently the Fedora rawhide Docker images carry an unusable /etc/dnf/dnf.conf file that we have to remove to get dnf doing anything.


jimbair (approved):
It's a shame about the codependency and conflict between DNF and DNF5, but this should work until we can clean it up later

Issue #1197, opened 2023-06-01
changedfiles: latest rpminspect is getting OOM-killed


msrb:
**Describe the bug**

It seems like 4e448f334b7217761f38a9e1676cc6fb3b6ba15e introduced some regression (in changedfiles inspection) that causes rpminspect to consume all available memory only to be OOM-killed immediately after that.

Versions before 4e448f334b7217761f38a9e1676cc6fb3b6ba15e don't exhibit this behavior.

The problem doesn't affect all tested components, but for example kernel is affected.

**To Reproduce**
```
$ /usr/bin/rpminspect -c /usr/share/rpminspect/redhat.yaml -T changedfiles --arch x86_64,noarch  --output=result.txt --verbose --release=el9 --profile=rhel-9 kernel-5.14.0-284.11.1.el9_2 52968362
...
=> kernel-devel-matched-5....4.11.1.el9_2.x86_64.rpm  [######################################################]
=> kernel-abi-stablelists-...4.11.1.el9_2.noarch.rpm  [######################################################]
=> kernel-doc-5.14.0-284.11.1.el9_2.noarch.rpm        [######################################################]
Skipping abidiff inspection...       skip
Skipping addedfiles inspection...    skip
Skipping annocheck inspection...     skip
Skipping arch inspection...          skip
Skipping badfuncs inspection...      skip
Skipping capabilities inspection...  skip
Killed

$ rpm -qa|grep rpminspect
rpminspect-data-centos-1.3-0.1.202303151248git53bfd31.fc38.noarch
rpminspect-data-fedora-1.11-0.1.202304251747gitcec0c97.fc38.noarch
rpminspect-data-redhat-1.8-0.1.202305250802gite1d03b4.fc38.noarch
librpminspect-1.12-0.1.202305311232git8c822b1.fc38.x86_64
rpminspect-1.12-0.1.202305311232git8c822b1.fc38.x86_64
```

My laptop has 32 gigs of memory.

**Expected behavior**
No OOM :)

**Environment:**
RHEL CI

dcantrell:
Westford labs are still down, so I can't talk to Brew.  Have to wait for that to come back.  In the interim I am testing with some Fedora kernels.

dcantrell:
Well, no.  Those are down too.  Fun.

dcantrell:
This one is taking longer to fix than I wanted it to.  I currently have it passing now, but there is a leak I am trying to track down.

dcantrell:
[...]

PR #1202, opened 2023-06-06
Restructure MIME handling and prevent crashes in 'changedfiles'

dcantrell:
A lot of changes here, but the main things are:

- Introduction of string_hash_t
- Caching of MIME type strings and storing pointers to the strings for each rpmfile_entry_t rather than a duplicated string
- Handling of libmagic giving us NULL
- Closing leaks that were (slooooooooooowly) found by valgrind

### DIFF
--- a/include/rpminspect.h
+++ b/include/rpminspect.h
@@ -296,9 +296,9 @@ void output_summary(const results_t *results, const char *dest, const severity_t
 int unpack_archive(const char *, const char *, const bool);
 
 /* magic.c */
-char *mime_type(const char *);
-char *get_mime_type(rpmfile_entry_t *);
-bool is_text_file(rpmfile_entry_t *);
+const char *mime_type(struct rpminspect *, const char *);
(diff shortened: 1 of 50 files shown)


jimbair (approved):
Lots of little changes but nothing looks out of sorts from here. Ideally a second +1 would be good given the size of the PR

evan-goode (approved):
Nice, does it look like the memory consumption issue described in https://github.com/rpminspect/rpminspect/issues/1197 is fixed in this PR? Or is there still more work to be done?

dcantrell:
> Nice, does it look like the memory consumption issue described in #1197 is fixed in this PR? Or is there still more work to be done?

There's still more work to do.  A full kernel build comparison running the changedfiles inspection balloons up to 108GB on my workstation.  But it no longer crashes and no longer has the kernel kill it.  I wanted to push this set now since it's basically ready (excepting the CI failures right now, which I am working on).

PR #1204, opened 2023-06-13
Portability fixes for packaging on NixOS

evan-goode:
I'm working on packaging rpminspect for NixOS (https://github.com/NixOS/nixpkgs/compare/master...evan-goode:nixpkgs:evan-goode/rpminspect) and ran into a couple things that I had to tweak.

1. For systems that don't have /usr/, the path to elf.h should be overrideable by an environment variable. Though who know whether any of pic_bits.sh will work/be relevant on NixOS :)
2. We should add additional include directories only if they exist on the system. Nix has its own way of telling Meson the right include paths to search.

### DIFF
--- a/lib/pic_bits.sh
+++ b/lib/pic_bits.sh
@@ -32,7 +32,9 @@
 # reloc types are in the R_<arch>_<type> #define's. The only case we (sort of)
 # care about where the arch's don't match is ia64: it's EM_IA_64 but R_IA64_*.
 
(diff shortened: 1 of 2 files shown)


dcantrell (changes requested):
Looks good to me, except please use 4 spaces for indentation in meson.build.

evan-goode:
Whoops, done.

dcantrell (approved):

PR #1221, opened 2023-07-17
[lib] Use execvp() rather than execvpe() in run_cmd()

dcantrell:
This is a portability change.  I have run in to some systems that have execvp() but not execvpe().  That's easy enough to deal with, so don't assume we have execvpe().

### DIFF
--- a/lib/runcmd.c
+++ b/lib/runcmd.c
@@ -30,7 +30,7 @@
 #endif
 
 /*
- * Given a command name, try to find it in the pATH.  Returns an
+ * Given a command name, try to find it in the PATH.  Returns an
  * allocated string that the caller is responsible for freeing.
  */
 char *find_cmd(const char *cmd)
--- a/lib/runcmd.c
+++ b/lib/runcmd.c
@@ -51,6 +51,13 @@ char *find_cmd(const char *cmd)
         return NULL;
     }
 
+    /* if cmd contains a '/', use as-is */
+    if (strchr(cmd, '/')) {
+        r = strdup(cmd);
+        assert(r != NULL);
+        return r;
+    }
+
     /* get the effective UID and GID */
     u = geteuid();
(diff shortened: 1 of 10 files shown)


lib/runcmd.c
```diff
@@ -22,6 +22,97 @@
...
 
+/* (pretend this program may build on Windows at some point) */
+#if defined(_WIN32) || defined(_WIN64) || defined(__CYGWIN__)
+#define PATH_ENV_SEP ";"
+#else
+#define PATH_ENV_SEP ":"
+#endif
+
+/*
+ * Given a command name, try to find it in the pATH.  Returns an
```
jimbair, line 33: Small nit; I assume it should be PATH

jimbair (approved):

lib/runcmd.c
dcantrell: Fixed.

evan-goode:
Another nit: don't look in PATH if `cmd` contains a `/`: https://git.musl-libc.org/cgit/musl/tree/src/process/execvp.c#n9

Soon glibc and musl will go extinct. I'd like to interject for a moment. What you're referring to as Linux, is in fact, rpminspect/Linux...

evan-goode (approved):
But that's minor, still approving.

dcantrell:
> Soon glibc and musl will go extinct. I'd like to interject for a moment. What you're referring to as Linux, is in fact, rpminspect/Linux...

Wait, what?

evan-goode:
I mean someday rpminspect will implement all of libc and we can switch over :)

dcantrell:
> I mean someday rpminspect will implement all of libc and we can switch over :)

Probably so.  It may even do that before it can send email[1].

[1] https://en.wikipedia.org/wiki/Jamie_Zawinski#Zawinski's_Law

PR #1222, opened 2023-07-17
[lib] Ignore missing optional commands in gather_diags()

dcantrell:
Some commands are optional, so if they error out trying to get information just skip listing them in the diagnostics output.

### DIFF
--- a/lib/diags.c
+++ b/lib/diags.c
@@ -192,85 +192,100 @@ string_list_t *gather_diags(struct rpminspect *ri, const char *progname, const c
 
     /* msgunfmt */
     tmp = run_cmd(&exitcode, ri->worksubdir, ri->commands.msgunfmt, "--version", NULL);
-    details = strsplit(tmp, "\n");
-    free(tmp);
 
-    entry = TAILQ_FIRST(details);
-    ver = strdup(entry->data);
-    list_free(details, free);
+    if (exitcode == 0) {
+        details = strsplit(tmp, "\n");
+        free(tmp);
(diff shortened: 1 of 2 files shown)


jimbair (approved):

lib/diags.c
```diff
@@ -192,85 +192,100 @@ string_list_t *gather_diags(struct rpminspect *ri, const char *progname, const c
...
+    if (exitcode == 0) {
+        details = strsplit(tmp, "\n");
+        free(tmp);
 
-    entry = calloc(1, sizeof(*entry));
-    assert(entry != NULL);
-    xasprintf(&entry->data, "%s", ver);
-    TAILQ_INSERT_TAIL(list, entry, items);
+        entry = TAILQ_FIRST(details);
+        ver = strdup(entry->data);
```
evan-goode, line 201: Will these segfault if, for some reason, the command exits 0 but doesn't print  any output?

dcantrell: It shouldn't segfault.  What happens is that optional commands will get a shell error that gets collected by gather_diags().  Like "flurplecmd: Command not found" with an exit code of 255.  So I'm checking for a non-zero exit code and making the jump to "you don't have this command".

I guess technically details could be empty even if the command exists, but in that case it would be a change in behavior of the command no longer printing version information or something like that.  I could guard against NULL pointer dereferencing and add in a fallback that just gives the path to the command or something like that.

evan-goode (approved):

PR #1223, opened 2023-07-19
[ci] Remove Rocky Linux from the Extra CI job collection

jimbair:


dcantrell (changes requested):
This looks good, but this commit should also remove osdeps/rocky8/ and osdeps/rocky9/

jimbair:
My mistake; I forgot to remove those as well. Updated.

dcantrell (approved):

Issue #1219, opened 2023-07-14
wrong permissions check


sharkcz:
**Describe the bug**
rpminspect reports a permissions check failure for the s390utils package
````
permissions:
------------
1) /var/log/ts-shell in s390utils-iucvterm on s390x carries unexpected mode 0770; expected mode 42770; requires inspection by the Security Team 

Result: BAD
Waiver Authorization: Security
````
see also https://bodhi.fedoraproject.org/updates/FEDORA-2023-73b4cfa106

the fedora rpminspect data contain (https://github.com/rpminspect/rpminspect-data-fedora/blob/main/fileinfo/fc39#L96C1-L96C60)
`drwxrws---          root       ts-shell   /var/log/ts-shell`
and the built `s390utils-iucvterm-2.28.0-1.fc39.s390x.rpm` contains
`drwxrws---    2 root     ts-shell                    0 Jul 11 02:00 /var/log/ts-shell`

So it's not clear why the check fails. But I have noticed that the rpm is extracted under `/var/tmp/rpminspect` with the ownership of the user running `rpminspect`, also the permissions are different (0750 for the `ts-shell` dir), so it's not the exact content as the rpm.

**To Reproduce**
`/usr/bin/rpminspect -T permissions -c /usr/share/rpminspect/fedora.yaml --release=fc39 --profile=rawhide s390utils-2.27.0-1.fc39 103318012`

**Expected behavior**
check passes

**Environment:**
 - whatever fedora-ci runs, but reproduced also on F-38/x86_64 and F-38/s390x and some other combos

sharkcz:
for the record, internal OSCI doesn't have such problem

dcantrell:
I bet we have mismatched vendor data.  I'll check on this.

PR #1224, opened 2023-07-20
Get file modes from the RPM header and not the cpio stream

dcantrell:
I was doing this for the owner and group, but not the modes.  I was just lifting that from the cpio stream.  But RPM keeps the file modes in the RPM header under the RPMTAG_FILEMODES array.  So read those and use that value for st.st_mode in extract_rpm().

Fixes: #1219


jimbair (approved):
Looks good from here =)

PR #1225, opened 2023-07-21
Update README.md and .mailmap

dcantrell:
Jim committed using his personal address and not work address, so map that.  Update the CentOS version list in the distro table.


evan-goode (approved):

jimbair:
Good catch; changed laptops and it went right off the rails. =)

PR #1232, opened 2023-08-09
[doc] Add .readthedocs.yaml

dcantrell:
Preparing for this change:
https://blog.readthedocs.com/migrate-configuration-v2/


jimbair (approved):
Seems fine to me, though I suspect we'll find out when it gets fully migrated.

dcantrell:
Agreed.  If anything, this is more of a placeholder for what is to come.  Full and complete docs are still on the to do list, which I save for lazy days.

PR #1245, opened 2023-08-30
[fix] Improve the shellsyntax noun/verb result messages

dcantrell:
Some of these were too generic or copied from another result type. Make these more specific so you know what you're looking at in the results.  These are only for the 'summary' output format.


evan-goode (approved):

jimbair (approved):
I always approve of better user messages =)

Issue #1246, opened 2023-09-01
Support multi-SRPM builds in rpmdeps inspection


jchecahi:
**Describe the bug**
Running `rpmdeps` inspection on modules can result in false positives regarding explicit package version requirements for shared libraries. This will happen because modules are generally a grouping of several packages, hence containing subpackages coming from different SRPMs. However, the inspection assumes that all the subpackages come from the same SRPM. Hence, it may infers that two subpackages belong to the same package when they don't.

Specific use case: llvm-toolset module. `clang-libs` subpackage from `clang` package needs `libLLVM-XX.so` (`XX` being LLVM major version). This is provided by `llvm-libs` subpackage from `llvm` package:
```
Subpackage clang-libs on aarch64 carries 'Requires: libLLVM-16.so()(64bit)' which comes
from subpackage llvm-libs but does not carry an explicit package version requirement.
Please add 'Requires: llvm-libs = %{version}-%{release}' to the spec file to avoid the
need to test interoperability between various combinations of old and new subpackages. 

Result: VERIFY
```
However rpminspect should not report this as a missing explicit requires, as it's fine to not have `clang-libs` tied to the same `%{version}-%{release}` as `llvm-libs`, because they come from different packages. Also the [packaging guidelines](https://docs.fedoraproject.org/en-US/packaging-guidelines/#_explicit_requires) advise to not add unnecessary explicit requires

**To Reproduce**
Using aarch64 only for faster testing:
```
rpminspect-redhat -v -T rpmdeps --arch aarch64 llvm-toolset-rhel8-8090020230803191529.b46abd14
```

**Expected behavior**
rpminspect should not report shared library dependencies from a subpackage as "missing explicit requires" when the shared library is provided by a subpackage coming from a different SRPM.

dcantrell:
I am unable to reproduce this problem with the current source, so I think this has been resolved.  The rpmdeps inspection passes now for the module build in this example and I do remember working on it in the past but perhaps I forgot to close the issue.

PR #1248, opened 2023-09-04
[fix] Prevent false positives in rpmdeps inspection for modules

jchecahi:
Ensure that, when requesting an explicit package requires for a shared library, both provider and requester packages come from the same SRPM.

This PR aims to prevent these false possitives described in #1246.

### DIFF
--- a/lib/inspect_rpmdeps.c
+++ b/lib/inspect_rpmdeps.c
@@ -206,6 +206,8 @@ static bool check_explicit_lib_deps(struct rpminspect *ri, Header h, deprule_lis
     bool result = true;
     const char *name = NULL;
     const char *arch = NULL;
+    const char *srpm = NULL;
+    const char *peer_srpm = NULL;
     rpmpeer_entry_t *peer = NULL;
     rpmpeer_entry_t *potential_prov = NULL;
     deprule_entry_t *verify = NULL;
@@ -236,6 +238,7 @@ static bool check_explicit_lib_deps(struct rpminspect *ri, Header h, deprule_lis
 
     name = headerGetString(h, RPMTAG_NAME);
     arch = get_rpm_header_arch(h);
+    srpm = headerGetString(h, RPMTAG_SOURCERPM);
 
     init_result_params(&params);
     params.waiverauth = WAIVABLE_BY_ANYONE;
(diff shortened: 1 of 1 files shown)


jchecahi:
Tested this with the same command described in issue #1246, the inspection no longer reports false possitives:
```
# ./build/src/rpminspect -c /usr/share/rpminspect/redhat.yaml -T rpmdeps --arch aarch64 llvm-toolset-rhel8-8090020230803191529.b46abd14
[...]
rpmdeps:                                                                                                                                                                                                                                                                        
--------                                                                                                                                                                                                                                                                        
Result: OK                                                                                                                                                                                                                                                                      
[...]
```

lib/inspect_rpmdeps.c
```diff
@@ -260,6 +263,14 @@ static bool check_explicit_lib_deps(struct rpminspect *ri, Header h, deprule_lis
                 continue;
             }
 
+            /*
+             * issue 1246: Prevent false positives when the library is provided
+             * by a subpackage coming from another SRPM.
+             */
+            peer_srpm = headerGetString(peer->after_hdr, RPMTAG_SOURCERPM);
+            if (peer_srpm && strcmp(peer_srpm, srpm)) {
```
dcantrell, line 271: I think this is a suitable workaround now until I can get better cross-RPM dependency handling for modules.  I would prefer that we also guard this test to ri->buildtype as KOJI_BUILD_MODULE.

dcantrell (changes requested):
I think this is a reasonable way to deal with dependencies that cross RPMs in modules for now, but I would like to restrict it to KOJI_BUILD_MODULE types.

lib/inspect_rpmdeps.c
jchecahi: Thanks for checking. That guard makes sense, I've updated the PR with the guard.

lib/inspect_rpmdeps.c
```diff
@@ -260,6 +263,15 @@ static bool check_explicit_lib_deps(struct rpminspect *ri, Header h, deprule_lis
                 continue;
             }
 
+            /*
+             * issue 1246: Prevent false positives when the library is provided
+             * by a subpackage coming from another SRPM.
+             */
+            peer_srpm = headerGetString(peer->after_hdr, RPMTAG_SOURCERPM);
```
dcantrell, line 270: Blank line after line 270, please.

lib/inspect_rpmdeps.c
```diff
@@ -260,6 +263,15 @@ static bool check_explicit_lib_deps(struct rpminspect *ri, Header h, deprule_lis
...
+            if (ri->buildtype == KOJI_BUILD_MODULE
```
dcantrell, line 271: Join lines 271 and 272, please.

lib/inspect_rpmdeps.c
```diff
@@ -260,6 +263,15 @@ static bool check_explicit_lib_deps(struct rpminspect *ri, Header h, deprule_lis
...
+                && peer_srpm && strcmp(peer_srpm, srpm)) {
+                continue;
+            }
```
dcantrell, line 274: Blank line after 274, please.

lib/inspect_rpmdeps.c
jchecahi: I also moved the comment immediately above the condition check, it makes more sense.

dcantrell (approved):

Issue #984, opened 2022-12-20
rpmdeps reports missing epoch for dep with explicit epoch


notroj:
**Describe the bug**
rpmdeps is reporting a missing epoch for a dependency which has an explicit epoch and is displayed with an explicit epoch in the error message:

`Missing epoch prefix on the version-release in 'Requires: httpd-core = 0:2.4.54-9.fc38' for mod_proxy_html on i686`

**To Reproduce**
```
/usr/bin/rpminspect -c /usr/share/rpminspect/fedora.yaml --workdir /var/ARTIFACTS/work-rpminspectprhh6lw7/rpminspect/discover/default-0/tests/workdir/ --format=json --output=/var/ARTIFACTS/work-rpminspectprhh6lw7/rpminspect/execute/data/rpminspect/data/result.json --verbose --release=fc38 --profile=rawhide httpd-2.4.54-6.fc38 95557571
```

**Expected behavior**
Either the real error is something more subtle than a missing epoch and is reported badly - or the report is a false positive?

**Environment:**
 - Fedora rawhide i686

dcantrell:
The problem here is rpminspect is not handling cases where you have an Epoch > 0 on a subpackage but not the main package.  I see httpd is doing that.  This should not be too difficult to fix up.

notroj:
Awesome, thanks David.

PR #1253, opened 2023-09-06
[fix] Handle subpackages with different Epoch values in rpmdeps

dcantrell:
Sometimes subpackages in one SRPM will carry different Epoch values that other subpackages or the main package.  This is allowed but the rpmdeps inspection was going with the old practice of "all subpackages will be on the same Epoch".  Now rpminspect will verify if an explicit Epoch value is needed in a deprule based on whether that named subpackage has a Epoch value greater than zero.  When no Epoch is specified, rpm assumes 0 and in those cases rpminspect does not care what the deprule carries in the version field.

Fixes: #984

### DIFF
--- a/lib/inspect_rpmdeps.c
+++ b/lib/inspect_rpmdeps.c
@@ -430,21 +430,20 @@ static bool check_explicit_lib_deps(struct rpminspect *ri, Header h, deprule_lis
 }
 
 /*
- * For packages with an Epoch > 0, check each dependency rule string
- * that uses the package version and release and ensure it comes
- * prefixed with the Epoch followed by a colon.
+ * For packages in a deprule that carry an Epoch > 0, make sure they
+ * are listed with the explicit Epoch value in the deprule.
  */
 static bool check_explicit_epoch(struct rpminspect *ri, Header h, deprule_list_t *afterdeps)
 {
     bool result = true;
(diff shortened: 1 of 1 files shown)


jimbair (approved):
Looks good; switching to pname (I assume package name?)  on line 455 seems fine, though I don't know if renaming arch and epoch to parch and pepoch would make it more or less clear.

But I see name used in the loop later so clearly it's functioning as intended. Mostly want to make sure it makes sense on the whole. But overall, looks good.

dcantrell:
> Looks good; switching to pname (I assume package name?) on line 455 seems fine, though I don't know if renaming arch and epoch to parch and pepoch would make it more or less clear.

I only renamed name to pname because I needed another string variable to hold RPMTAG_NAME.  Now pname holds RPMTAG_NAME for the package we are looking at the deprules for.  Later, name is used to get the subpackage name which is used to find the header of that package.

> But I see name used in the loop later so clearly it's functioning as intended. Mostly want to make sure it makes sense on the whole. But overall, looks good.

The reporting will show what package needs a dependency change in it, which is where pname is used.  Then the deprule string (drs) is displayed.  'name' is only used in the loop over each deprule as it finds that subpackage header to gather both the arch and epoch.

PR #1260, opened 2023-09-14
[lib] Expand the desktop inspection with more configuration opts

dcantrell:
Allow the desktop configuration block to allow users to list paths to .desktop files to skip either the Exec= check or Icon= check.  This allows the desktop inspection to still run desktop-file-validate but skip the Exec and Icon checks when the subpackages do not carry either the application or icon file.


jimbair (approved):
Looks good from over here

Issue #1040, opened 2023-01-25
No ownership and mode checks performed for binaries with capabilities


thoger:
**Is your feature request related to a problem? Please describe.**
This request is similar to what's noted in #1039 for setuid/setgid binaries, but for binaries with file capabilities.  For those, we should perform ownership and mode checks as well, not only capabilities list check.

However, unlike the problem for suid/sgid, this is also a design issue, not only implementation issue.  The capabilities allow-list format only contains the following information: package, path, and capabilities.  See: https://github.com/rpminspect/rpminspect/blob/v1.10/data/capabilities/GENERIC

The ownership and mode check is desired in cases like wireshark's /usr/bin/dumpcap, which is owned by root:wireshark and with mode 750 (i.e. only allowing wireshark group users to execute it).

**Describe the solution you'd like**
I'm open to discussion how to best address this - whether to require more information in capabilities allow-list, try to merge fileinfo and capabilities lists, or something else.

I think the current work around is to add any binary with capabilities to both fileinfo and capabilities lists.  The listing in fileinfo would ensure (once #1039 is fixed) ownership and mode checks.  This approach has drawbacks:

- The list of two files makes it likely to have incompletely or our-of-sync configuration.
- fileinfo listing is currently only required for suid/sgid, but not for capabilities.
- fileinfo is already meant to be used for other purposes - to allow-list not privileged binaries in bin_paths with different ownership than root:root.  Note that I'm not aware of any particular example of such binary and its listing in a fileinfo file.  For these not privileged binaries, there is no need to require security review.  But given that this case is not seen in the wild, the current status-quo seems acceptable.  Noting this explicitly so it can be kept in mind if designing any new allow-listing solution.

dcantrell:
At this point I don't want to change the formats of the fileinfo or capabilities files.  They exist in the wild now and I don't want to have to go and change them and ensure backwards compatibility exists.

I think having ownership info recorded in fileinfo for files that also need entries in the capabilities file is the way to go.  You are right, I currently only use it for a few checks, but there's no reason it can't also be used for ownership checks in the capabilities inspection.

Thoughts?

dcantrell:
So I'm looking at this one again and I don't see the concern.  The ownership inspection will check the ownership information, permissions will check permissions, and capabilities will check capabilities of a file.  I don't understand why the capabilities inspection needs to also check ownership and permissions -or- why that data needs to be in the same configuration file in the vendor data package.

[...]

PR #1266, opened 2023-09-25
Check mode+ownership in capabilities inspection

evan-goode:
Resolves https://github.com/rpminspect/rpminspect/issues/1040

The capabilities inspection currently does not check file permissions and ownership, so a world-writable binary with capabilities could pass the inspection. To catch such cases, the capabilities inspection should also check file modes and ownership, even when the permissions and ownership inspections themselves are not being run.

The tests still need some work.


dcantrell (changes requested):
Overall this is what I was wanting to do.  I only have a handful of requests:

1. For consistency in the source, I'd like the check_mode() function in permissions.c and the check_ownership() in ownership.c.  I'm going to get lost trying to find check_ownership() in the future.  Alternatively, a single file named attributes.c or something like that.  Maybe fileattrs.c, but there are a number of source files that begin with "file" already.
2. I would prefer check_mode() be named check_permissions().  Throughout rpminspect, I refer to the mode_t bits as "permissions" since that's the common end user language.  Also mode_t and mode are used throughout the standard C library.
3. The tests need "make flake8" run and the problems fixed.
4. The tests also need "make black" run and the problems fixed.

For the Python checks, I have a pre-push hook that I lifted from the Tor project and use in my local repo.  I'll add it to utils/git-hooks and you can drop it in your ~/.git/hooks subdirectory.  It runs flake8, black, and shellcheck before pushing changes you've committed which is handy because those checks will run when you make a PR.

The rawhide build and test run also failed and it looks to be a bug in lib/init.c.  It's failing to initialize the ri->ignore global ignore list.  Now the test suite uses data/generic.yaml as the basis for the rpminspect config file and it makes some path changes (which you can see in test/baseclass.py.  But yesterday I commented out everything in the global ignore list, but I left the block name in place.  This is probably the source of the bug.  I believe rpminspect should be able to handle an empty block like that.  If I have time today, I'll try to get that fixed.  Once that's fixed, the global ignore block in data/generic.yaml should be commented out.

evan-goode:
> Overall this is what I was wanting to do. I only have a handful of requests:
> 
>     1. For consistency in the source, I'd like the check_mode() function in permissions.c and the check_ownership() in ownership.c.  I'm going to get lost trying to find check_ownership() in the future.  Alternatively, a single file named attributes.c or something like that.  Maybe fileattrs.c, but there are a number of source files that begin with "file" already.

OK, moved check_ownership to ownership.c.

> 
>     2. I would prefer check_mode() be named check_permissions().  Throughout rpminspect, I refer to the mode_t bits as "permissions" since that's the common end user language.  Also mode_t and mode are used throughout the standard C library.

Done
> 
>     3. The tests need "make flake8" run and the problems fixed.
>     4. The tests also need "make black" run and the problems fixed.

Done

> 
> 
> For the Python checks, I have a pre-push hook that I lifted from the Tor project and use in my local repo. I'll add it to utils/git-hooks and you can drop it in your ~/.git/hooks subdirectory. It runs flake8, black, and shellcheck before pushing changes you've committed which is handy because those checks will run when you make a PR.
>
Maybe a [.pre-commit-config.yaml](https://pre-commit.com/) would be good to make it more automatic? That's what we use for DNF 5.

> The rawhide build and test run also failed and it looks to be a bug in lib/init.c. It's failing to initialize the ri->ignore global ignore list. Now the test suite uses data/generic.yaml as the basis for the rpminspect config file and it makes some path changes (which you can see in test/baseclass.py. But yesterday I commented out everything in the global ignore list, but I left the block name in place. This is probably the source of the bug. I believe rpminspect should be able to handle an empty block like that. If I have time today, I'll try to get that fixed. Once that's fixed, the global ignore block in data/generic.yaml should be commented out.

I think I see the reason my test fails, the ownership checks are getting skipped when `INSPECT_OWNERSHIP` is not set on `ri->tests`. I'll fix it.

dcantrell (approved):

Issue #1268, opened 2023-09-26
Composed messages can not be properly translated


goeranu:
After [line 554 of inspect_patches.c](https://github.com/rpminspect/rpminspect/blob/550fc49bb834c56b341e7a81c834b2342e055be0/lib/inspect_patches.c#L554) there are three print statements that compose a message that optionally adds an "s" to a word if there are more than one line/file, and also might insert "as many as".

Composing messages in that way makes them impossible to translate properly. "S" is only a plural suffix in some languages obviously. Furthermore, many languages have other plural distinctions than simple singular/plural like English. In that case, Swedish which I'm translating into has the same pattern as English, but many languages distinguish more cases.

The correct way to handle this is to use the `ngettext` function, where you give the English singular and plural and also the number, and then the function can select between serveral messages different for each target language.

dcantrell:
Thank you!  I am definitely not an expert on the translation functions, but I wanted to set things up in the code so that strings could be translated.  I will prepare a patch as you describe for inspect_patches.c and let me know if that looks correct to you.  If it does, I can merge it and get the translation template updated.  There are probably many other examples like this in the code, so please do point them out so I can fix them.

goeranu:
I'm sure you can google it yourself, but in case you would like here is a [pointer](https://www.gnu.org/software/gettext/manual/html_node/Plural-forms.html) to the documentation around `ngettext`.

dcantrell:
Yep, I found it.

dcantrell:
Does this pull request look correct to fix the translation problem with these strings?

goeranu:
Yes, it looks good to me!

PR #1270, opened 2023-09-26
Use N_() for result strings in the patches inspection

dcantrell:
The reporting in the patches inspection notes how many file(s) and/or line(s) a particular patch touches.  Handle the plural form of the result strings with the ngettext() function so that other languages can be translated correctly.
    
Fixes: #1268

### DIFF
--- a/include/rpminspect.h
+++ b/include/rpminspect.h
@@ -67,7 +67,7 @@ extern volatile sig_atomic_t terminal_resized;
 #define N_(MSGID, MSGID_PLURAL, N) ngettext((MSGID), (MSGID_PLURAL), (N))
 #else
 #define _(MSGID) (MSGID)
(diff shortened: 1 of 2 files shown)


dcantrell:
@goeranu will this change correct the translation problem you were reporting?

goeranu:
This looks good to me. I haven't actually built an ran it, but it looks just how I would hope.

(Maybe you want to chose the plural case for the definition of `N_` when gettext is **not** available? A matter of taste.)

dcantrell:
I did not know what a common or preferred style is when ngettext() is not available.  I guess defaulting to the plural one makes sense.  I will change that macro and then merge the PR.  Thanks!

goeranu:
I don't know there really is any generally agreed preferred style when you don't have `ngetttext()` etc. It's just my personal observation that I more often see things like "1 files" than I see "2 file".

Thank you for fixing!

PR #1275, opened 2023-10-02
Set up pre-commit

evan-goode:
pre-commit is a modern tool for managing pre-commit hooks to run linters, formatters, etc: https://pre-commit.com/

You can install pre-commit (e.g. with `pip install pre-commit`) and then run `pre-commit install` in your clone of rpminspect to set up the pre-commit hooks.

See commit messages.


.github/workflows/pre-commit.yml
```diff
@@ -0,0 +1,14 @@
+name: pre-commit
```
dcantrell, line 1: Please use 4 space indentation for YAML files.  They're already nearly unreadable, I need 4 space spaces in order to comprehend them.

.github/workflows/python.yml
```diff
@@ -1,31 +0,0 @@
-name: Python Checks
```
dcantrell, line 1: Is this workflow being removed in favor of using the pre-commit method?  If so, I would prefer to keep this one here in the classic GHA sense in case contributors are not relying on the pre-commit piece.

.pre-commit-config.yaml
```diff
@@ -0,0 +1,16 @@
+repos:
+  - repo: https://github.com/pre-commit/pre-commit-hooks
+    rev: v4.4.0
```
dcantrell, line 3: I am wary of explicit version requirements for CI stuff because it ends up being something to chase.  Is there a way to indicate "follow stable" or something like that?

.pre-commit-config.yaml
```diff
@@ -0,0 +1,16 @@
...
+    hooks:
+      - id: check-yaml
+      - id: end-of-file-fixer
+      - id: trailing-whitespace
+  - repo: https://github.com/psf/black
+    rev: 23.9.1
```
dcantrell, line 9: Same as rev comment above.

.pre-commit-config.yaml
```diff
@@ -0,0 +1,16 @@
...
+    hooks:
+      - id: black
+  - repo: https://github.com/PyCQA/flake8
+    rev: 6.1.0
```
dcantrell, line 13: Same as rev comment above.

.pre-commit-config.yaml
```diff
@@ -0,0 +1,16 @@
+repos:
+  - repo: https://github.com/pre-commit/pre-commit-hooks
+    rev: v4.4.0
+    hooks:
+      - id: check-yaml
+      - id: end-of-file-fixer
+      - id: trailing-whitespace
```
dcantrell, line 7: The trailing-whitespace thing is going to be a problem because some of the test suite data deliberately relies on this.  Maybe adding an ignore list explicitly, if that's possible.

CC-BY-4.0.txt
```diff
@@ -393,4 +393,3 @@ the avoidance of doubt, this paragraph does not form part of the
 public licenses.
 
 Creative Commons may be contacted at creativecommons.org.
-
```
dcantrell, line 396: Do not modify the license text files, those come as-is from elsewhere.  If these need to be ignored by the pre-commit hooks, then we should do that.

CODE_OF_CONDUCT.md
```diff
@@ -138,4 +138,3 @@ are available at
 [Mozilla CoC]: https://github.com/mozilla/diversity
 [FAQ]: https://www.contributor-covenant.org/faq
 [translations]: https://www.contributor-covenant.org/translations
-
```
dcantrell, line 141: Do not modify this file, it comes from elsewhere.  pre-commit should ignore it.

MIT.txt
```diff
@@ -17,9 +17,9 @@ Content-Type: text/plain; charset=utf-8
 Content-Disposition: inline
 
 > Just wanted to say thanks for the detailed write up here:
->
```
dcantrell, line 20: This file is not to be modified.

libxdiff/COPYING
```diff
@@ -55,7 +55,7 @@ modified by someone else and passed on, the recipients should know
 that what they have is not the original version, so that the original
 author's reputation will not be affected by problems that might be
 introduced by others.
-
```
dcantrell, line 58: Do not modify the COPYING file in libxdiff.  It's an as-is copy from FSF.

libxdiff/README
```diff
@@ -1,12 +1,12 @@
 = libxdiff-1.0 =
 
-This is a merged libxdiff taken from the forks in git and libgit2.
```
dcantrell, line 3: Also don't want to modify the libxdiff README.

test/test_config.py
```diff
@@ -31,8 +31,8 @@
 #freebsd
 #openbsd
 #netbsd
-linux
```
dcantrell, line 34: These lines are required for the test cases in test_config.py

test/test_doc.py
```diff
@@ -20,7 +20,7 @@
 """
 
 readme_with_whitespace_changes = """This is a README file.
-With multiple lines.
```
dcantrell, line 23: This line is required for a test_doc.py test case.

test/test_patches.py
```diff
@@ -158,19 +158,19 @@
 -    return x + y;
 +    return (x + y);
  }
-
```
dcantrell, line 161: These lines need to be left as-is in test_patches.py.  It is a patch-compatible formatted diff that works across Linux and various BSD systems.

dcantrell (changes requested):
Left comments through-out.  Overall I like the idea of the pre-commit thing.  But since it requires installing a pip module, I would still like to have the regular Python GHA jobs for flake8 and black.

The license files and text files in libxdiff should not be modified.  I don't know if pre-commit can carry an ignore list.

And the whitespace changes in the test scripts should be left there because they are in place specifically to check how rpminspect handles whitespace in those cases.

Lastly, please use 4 spaces to indent and align YAML files.  I can't read it in my advanced age otherwise.

.github/workflows/pre-commit.yml
evan-goode: Done

.pre-commit-config.yaml
evan-goode: There actually isn't a (easy) way to do this by design, https://pre-commit.com/#using-the-latest-version-for-a-repository. In general I think it is better to keep the versions of pinned for the sake of toolchain reproducibility. There is `pre-commit autoupdate` to update the `rev`s there to the latest versions, so really they are akin to entries in a lockfile.

evan-goode:
I just took out the line ending/whitespace checks. I've noticed git is pretty good about warning me on trailing whitespace, it's probably more trouble than it's worth to lint that.

.github/workflows/python.yml
dcantrell: Oh, ok.  That's my misunderstanding.  In that case, kill it.

.pre-commit-config.yaml
dcantrell: Bleh, ok.  I like the idea of having it track whatever that project is calling their stable release or release or whatever rather than experiencing bit rot, but maybe there's no way around that.

dcantrell:
> I just took out the line ending/whitespace checks. I've noticed git is pretty good about warning me on trailing whitespace, it's probably more trouble than it's worth to lint that.

Agreed.  git has improved with that over the years.  Everything else looks fine.  Go ahead and drop the old python workflow since I misunderstood the pre-commit execution phase.

evan-goode:
> With the removal of tox.ini, the behavior of "make flake8" changes. Maybe we can remove most of what is in tox.ini, but I would like to keep enough present so that I can continue to run "make flake8" locally. I think it really just needs the [flake8] block and the rest can go.

Gotcha, done.

dcantrell (approved):

PR #1276, opened 2023-10-02
FreeBSD CI fixes and skip test_capabilities tests if rpm lacks %caps

dcantrell:
FreeBSD CI job needed some updates.

There were some tests in test_capabilities.py that were not guarded by the skipUnless check to see that rpm has %caps support.


evan-goode (approved):
Ah yep thanks!

PR #1280, opened 2023-10-10
CI fixes

dcantrell:
Fix the currently non-working CI jobs so that we get everything passing.  These updates were mostly centered around fixing the post.sh scripts to account for software changes.


jimbair (approved):
Looks good from here

PR #1282, opened 2023-10-16
[ci] Ensure the Slackware CI job builds the latest release of clamav

dcantrell:
Another instance of me missing 'sort -V' when looking at version numbers.


jimbair (approved):
Slackware; a classic.

PR #1283, opened 2023-10-16
[bld] Allow vendor data dir and work dir to be set at build time

dcantrell:
Previously the vendor data directory and the working directory were hardcoded and defined in the constants.h file.  The vendor data directory was set to /usr/share/rpminspect and the working directory was set to /var/tmp/rpminspect.  Those are still the defaults, but allow those locations to change at build time.

### DIFF
--- a/include/constants.h
+++ b/include/constants.h
@@ -84,11 +84,16 @@
 #define SOFTWARE_NAME "librpminspect"
 
 /**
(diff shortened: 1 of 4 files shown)


jimbair (approved):

evan-goode (approved):
Lovely

PR #1287, opened 2023-10-25
[fix] Recognize source RPM packages in subpackages test

ppisar:
If a subset of architectures including "src" was passed to rpminspect (e.g. -a x86_64,noarch,src), a subpackages test would fail if the source packages were built on different architectures (e.g. aarch64 and x86_64):

    Skipping specname inspection...      skip
    Running subpackages inspection...    FAIL
    Skipping symlinks inspection...      skip

The cause is that a source RPM package quotes an architecture where it was built in RPMTAG_ARCH tag, while rpminspect expected "src".

This patch fixes it by special-casing source packages.

In addition, if the source package was built on an unnamed (-a) architecture, the test logic reported an error without any details. This patch does not address this underreporting issue. A possible approach is moving "result = false" assignements into the if(allowed_arch()) branches in inspect_subpackages() function.

<https://centos.softwarefactory-project.io/zuul/t/centos/build/cb19599de10841c28ea3afb3ec2f1e7a>

### DIFF
--- a/lib/inspect_subpackages.c
+++ b/lib/inspect_subpackages.c
@@ -38,13 +38,13 @@ bool inspect_subpackages(struct rpminspect *ri)
     TAILQ_FOREACH(peer, ri->peers, items) {
         if (peer->before_hdr) {
             entry = calloc(1, sizeof(*entry));
-            xasprintf(&entry->data, "%s %s", headerGetString(peer->before_hdr, RPMTAG_NAME), headerGetString(peer->before_hdr, RPMTAG_ARCH));
+            xasprintf(&entry->data, "%s %s", headerGetString(peer->before_hdr, RPMTAG_NAME), get_rpm_header_arch(peer->before_hdr));
             TAILQ_INSERT_TAIL(before_pkgs, entry, items);
         }
 
         if (peer->after_hdr) {
             entry = calloc(1, sizeof(*entry));
-            xasprintf(&entry->data, "%s %s", headerGetString(peer->after_hdr, RPMTAG_NAME), headerGetString(peer->after_hdr, RPMTAG_ARCH));
+            xasprintf(&entry->data, "%s %s", headerGetString(peer->after_hdr, RPMTAG_NAME), get_rpm_header_arch(peer->after_hdr));
             TAILQ_INSERT_TAIL(after_pkgs, entry, items);
         }
     }


lib/inspect_subpackages.c
```diff
@@ -9,6 +9,17 @@
...
 #include "rpminspect.h"
 
+/*
+ * Fotmat an RPM package name and an architecture into a space-separated string.
+ * A source package is denoted with "src" architecture.
+ */
+static void format_subpackage_name_arch(char **data, Header hdr) {
+    xasprintf(data, "%s %s",
+        headerGetString(hdr, RPMTAG_NAME),
+        headerIsSource(hdr) ? "src" : headerGetString(hdr, RPMTAG_ARCH)
```
dcantrell, line 19: I had done this for other inspections too, I guess I just didn't use it here when I should have.  The function is get_rpm_header_arch() in rpm.c.  That can replace line 19.

lib/inspect_subpackages.c
```diff
@@ -38,13 +49,13 @@ bool inspect_subpackages(struct rpminspect *ri)
     TAILQ_FOREACH(peer, ri->peers, items) {
         if (peer->before_hdr) {
             entry = calloc(1, sizeof(*entry));
-            xasprintf(&entry->data, "%s %s", headerGetString(peer->before_hdr, RPMTAG_NAME), headerGetString(peer->before_hdr, RPMTAG_ARCH));
```
dcantrell, line 41: This line could probably stay by just replacing the headerGetString for RPMTAG_ARCH with a call to get_rpm_header_arch().  The format_subpackage_name_arch() function could be removed.

lib/inspect_subpackages.c
```diff
@@ -38,13 +49,13 @@ bool inspect_subpackages(struct rpminspect *ri)
...
+            format_subpackage_name_arch(&entry->data, peer->before_hdr);
             TAILQ_INSERT_TAIL(before_pkgs, entry, items);
         }
 
         if (peer->after_hdr) {
             entry = calloc(1, sizeof(*entry));
-            xasprintf(&entry->data, "%s %s", headerGetString(peer->after_hdr, RPMTAG_NAME), headerGetString(peer->after_hdr, RPMTAG_ARCH));
```
dcantrell, line 47: Same as above.

dcantrell (changes requested):
Thanks for the catch here.  I would prefer the patch change the original xasprintf() lines so it calls get_rpm_header_arch() as the last parameter rather than adding format_subpackage_name_arch().

dcantrell (approved):

PR #1288, opened 2023-10-26
Support new license DB format in the license inspection

dcantrell:
The fedora-license-data project wants to change the JSON data format:

https://gitlab.com/fedora/legal/fedora-license-data/-/issues/387

The format of the config file is effectively an API.  There's no versioning and there's no way to tell what config file format you are looking at.  I personally view this as an inappropriate change.  However, I don't control fedora-license-data and that was the whole point of getting that project going.

This massive patch set splits out some JSON handling from init.c in public API so it can be used by the license inspection.  I have modified the license inspection to try reading license db blocks in the new format first and if that fails, fall back on the old format.  Without any versioning information obtainable from the config file, there does not appear to be a clean way to do it otherwise.  Both config file formats exist in the wild, so supporting both while favoring the new format makes the most sense to me.

I have added some test cases as well to test the new db format in the license inspection.

During this work I also ran in to some test suite failures on Fedora and the annocheck inspection, so that's why there are commits to test_annocheck.py.

Sadly, this is not the last license inspection update.  The next patch (simpler) is to honor the forthcoming exceptions array in the db.  This is to handle cases where fedora-license-data records a license as not-allowed for Fedora, oh except for the things we consider exceptions.

### DIFF
--- a/include/rpminspect.h
+++ b/include/rpminspect.h
@@ -24,6 +24,7 @@
 #include "results.h"
 #include "output.h"
 #include "readelf.h"
+#include "parser.h"
 
 #ifndef _LIBRPMINSPECT_RPMINSPECT_H
 #define _LIBRPMINSPECT_RPMINSPECT_H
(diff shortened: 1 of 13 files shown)


jimbair (approved):
Nothing huge to add; I appreciate the commits to help break up the work that was done. I also appreciate that the change allows the old and new formats to work alongside each other given the lack of a numbered API between rpminspect and fedora-license-data.

PR #1302, opened 2023-11-15
[bld] Use SPDX-License-Identifier in regress/ scripts

dcantrell:


evan-goode (approved):
Any reason why one file has the year and the other doesn't?

dcantrell:
> Any reason why one file has the year and the other doesn't?

Because I forgot to delete it.

Issue #1295, opened 2023-11-09
rpminspect warns about all (?) symlinks in packages


keszybz:
https://artifacts.dev.testing-farm.io/0f49e367-3397-4fd8-8118-ce5bac36e2fd/

> `INFO` symbolic link /usr/lib/environment.d/99-environment.conf became a dangling symbolic link in systemd on i686 `Not Waivable`

It is a symlink to /etc/enviornment which is provided by `setup.rpm` which is required by `filesystem.rpm`, which means that it's always available.

> `VERIFY` regular file /usr/lib/systemd/systemd-cryptsetup became a symbolic link (to ../../bin/systemd-cryptsetup) in systemd-udev on i686 `Anyone`
> `VERIFY` regular file /usr/lib/systemd/ukify became a symbolic link (to ../../bin/ukify) in systemd-ukify on noarch `Anyone`

Those are just regular relative symlinks to a file in the same package. Maybe it doesn't understand relative symlinks?

dcantrell:
rpminspect should be able to see that absolute to relative symlinks still point to the same thing.  Still investigating, but this looks like a bug on my end.

PR #1307, opened 2023-11-28
Correctly resolve relative symlinks across subpackages

dcantrell:
The symlinks inspection was only checking the current subpackage for relative symlink resolution, but packages are allowed to have relative symlinks to other subpackages.  This patch fixes the symlinks inspection to correctly handle that case and report at the correct level (INFO, VERIFY, BAD).

Fixes: #1295

### DIFF
--- a/lib/inspect_symlinks.c
+++ b/lib/inspect_symlinks.c
@@ -22,6 +22,60 @@
 #include <err.h>
 #include "rpminspect.h"
 
+/*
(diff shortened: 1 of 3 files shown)


evan-goode (approved):
Looks good from here!

jimbair (approved):
Looks logic and such seems good to me, though I suspect some testing will confirm if any edge cases surprise us

Issue #1231, opened 2023-08-03
rpminspect complains about missing depedency that is provided transitively


keszybz:
> Subpackage systemd-tests on i686 carries 'Requires: libsystemd.so.0' which comes from subpackage systemd-libs but does not carry an explicit package version requirement. Please add 'Requires: systemd-libs = %{version}-%{release}' to the spec file to avoid the need to test interoperability between various combinations of old and new subpackages.

> Suggested remedy:
> Add the indicated explicit Requires to the spec file for the named subpackage. Subpackages depending on shared libraries in another subpackage must carry an explicit 'Requires: SUBPACKAGE_NAME = %{version}-%{release}' in the spec file.

The dependency is provided transitively:
```
$ rpm -qR systemd-tests|rg '^systemd'
systemd(x86-64) = 254~rc2-1.fc38
$ rpm -qR systemd|rg '^systemd-libs' 
systemd-libs = 254~rc2-1.fc38
```

It is a fairly common pattern to have such transitive deps between subpackages. I think rpminspect could figure out if a dependency is satisfied transitively among subpackages.

Also, I think the dependency needs to be archful. I.e. it should include `(%_isa)`. In systemd, the systemd subpackage doesn't have this, which is probably an error. The advice given by rpminspect should include it, because libraries by definition are compiled code and archful.

msrb:
Hmm :thinking: I'd argue that it is a good practice to list the dependencies explicitly. Do you see any problems or disadvantages with the explicit listing?

keszybz:
It's not that an explicit listing would be bad. But it might be unnecessary, and it is not an error to not have it. This warning is about subpackages of a single package, maintained together, and it's entirely reasonable for the maintainer to make use of transitive dependencies like this.   If this was between different packages, then I'd argue that it should always be declared explicitly, because the other package may rearrange components at any time. But when it's all in a single spec file, the maintainer fully controls all the subpackages. In particular, the maintainer might want to not add those dependencies explicitly to now blow up the number of dependencies.

In general, this is another of those cases where some advice might be useful in some cases, but it's very hard to figure out programatically if it applies. Rpminspect presents this as certain.  I would very much prefer for rpminspect to not print the warning at all, so that we can get to point where warnings from rpminspect have no false positives, or almost no false positives, and then maintainers are motivated to fix the issues that are reported.

dcantrell:
[...]

PR #1309, opened 2023-11-30
[fix] In the symlinks inspection, honor transitive Requires

dcantrell:
A package with automatic shared library dependencies usually require an explicit Requires on the %{name} = %{version}-%{release}.  This is done to ensure subpackages don't drift from the main package when performing upgrades.  However, it is also possible to have those explicit Requires handled by transitive Requires.  That is having an explicit Requires on a package that then itself has an explicit Requires on the package you need.  This commit implements that check when looking for explicit Requires and treats it as satisifed.

Fixes: #1231

### DIFF
--- a/lib/inspect_rpmdeps.c
+++ b/lib/inspect_rpmdeps.c
@@ -433,43 +433,45 @@ static bool check_explicit_lib_deps(struct rpminspect *ri, Header h, deprule_lis
                         }
                     }
 
-                    collect = true;
+                    if (transitive && !TAILQ_EMPTY(transitive)) {
+                        collect = true;
 
-                    while (collect) {
-                        TAILQ_FOREACH(peer, ri->peers, items) {
-                            tn = headerGetString(peer->after_hdr, RPMTAG_NAME);
-                            collect = false;
+                        while (collect) {
+                            TAILQ_FOREACH(peer, ri->peers, items) {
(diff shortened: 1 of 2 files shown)


lib/inspect_rpmdeps.c
```diff
@@ -378,6 +406,70 @@ static bool check_explicit_lib_deps(struct rpminspect *ri, Header h, deprule_lis
...
+                                if (!strcmp(tn, entry->data)) {
+                                    TAILQ_FOREACH(verify, peer->after_deprules, items) {
+                                        if (verify->type == TYPE_REQUIRES
+                                            && *verify->requirement != '/'
+                                            && !strprefix(verify->requirement, SHARED_LIB_PREFIX) && !strstr(verify->requirement, SHARED_LIB_SUFFIX)) {
+                                            isareq = remove_isa_substring(verify->requirement);
+                                            assert(isareq != NULL);
+
+                                            if (is_subpackage(ri->peers, isareq) && !list_contains(transitive, isareq)) {
+                                                transitive = list_add(transitive, isareq);
```
evan-goode, line 453: Could you break out of the `TAILQ_FOREACH` here if the `pn` is found instead of continuing the breadth-first search?

evan-goode (commented):
Looks good, do you think this deserves a test?

PR #1320, opened 2024-01-04
[bld] (Requires|Recommends): /usr/bin/annocheck -> annobin-annocheck

dcantrell:
Specify the package name rather than the executable for the dependency.  Path dependencies do not work with DNF 5/


evan-goode:
Unless I'm mistaken, filepath dependencies in `/usr/bin`, `/usr/sbin`, and `/etc` still work: https://docs.fedoraproject.org/en-US/packaging-guidelines/#_file_and_directory_dependencies. But this is still a good change.

PR #1322, opened 2024-01-11
[lib] Adjust _is_debug_rpm_helper() for other distributions

dcantrell:
On some systems this function would return false when looking at a debuginfo package that meets the criteria.

### DIFF
--- a/lib/rpm.c
+++ b/lib/rpm.c
@@ -519,9 +519,6 @@ static bool _is_debug_rpm_helper(Header hdr, const char *provide, const char *su
         if (strsuffix(n, suffix)) {
             rn = true;
         }
-    } else {
-        /* caller did not specify a suffix to check */
-        rn = true;
     }
 
     /* start new header transactions */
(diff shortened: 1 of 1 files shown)


evan-goode (approved):
Just so I understand, this to correctly count packages as debuginfo packages if they (1) are named `*-debuginfo` but do not provide `debug-info(build-id)` or (2) do not end in `-debuginfo` but do provide `debug-info(build-id)`?

dcantrell:
> Just so I understand, this to correctly count packages as debuginfo packages if they (1) are named `*-debuginfo` but do not provide `debug-info(build-id)` or (2) do not end in `-debuginfo` but do provide `debug-info(build-id)`?

Yeah, so there's no real required standard on identifying debuginfo packages.  On RH-derived systems we add "-debuginfo" to the RPMTAG_NAME.  But an autodep is generated for the debug-info(build-id) which is a more definitive way to suss out a debuginfo package, but is still not guaranteed everywhere.  So I try to do my best here.

PR #1323, opened 2024-01-15
[ci] Use actions/checkout@v4 for the git check part in CI jobs

dcantrell:
There's a new one and I hope it has fixed the safe.directory workaround.


evan-goode (approved):

jimbair (approved):
Higher numbers are generally better!

PR #1325, opened 2024-01-15
More CI and test case fixes

dcantrell:
The bit rot is real.


jimbair (approved):
Glad to see gentoo hanging in there

Issue #1329, opened 2024-01-29
segfault in the unicode inspection


msrb:
**Describe the bug**

rpminspect crashes in the unicode inspection when running on a specific build.

```
$ rpm -qa|grep rpminspect
rpminspect-data-centos-1.4-1.fc39.noarch
rpminspect-data-fedora-1.13-0.1.202312141553git52f36ba.fc39.noarch
rpminspect-data-redhat-1.9-0.1.202401171017gitc1131bb.fc39.noarch
librpminspect-2.0-0.1.202401241824gitd180976.fc39.x86_64
rpminspect-2.0-0.1.202401241824gitd180976.fc39.x86_64
```

**To Reproduce**

```
$ rpminspect-redhat -v -T unicode vim-9.1.016-1.el10+2
Downloading RPM build vim-9.1.016-1.el10+2
=> vim-9.1.016-1.el10+2.src.rpm                       [######################################################]
...
=> vim-common-9.1.016-1.el10+2.s390x.rpm              [######################################################]
Skipping abidiff inspection...       skip
Skipping addedfiles inspection...    skip
Skipping annocheck inspection...     skip
Skipping arch inspection...          skip
Skipping badfuncs inspection...      skip
Skipping capabilities inspection...  skip
Skipping changedfiles inspection...  skip
Skipping changelog inspection...     skip
Skipping config inspection...        skip
Skipping debuginfo inspection...     skip
Skipping desktop inspection...       skip
Skipping disttag inspection...       skip
Skipping doc inspection...           skip
Skipping dsodeps inspection...       skip
Skipping elf inspection...           skip
Skipping emptyrpm inspection...      skip
Skipping files inspection...         skip
Skipping filesize inspection...      skip
Skipping javabytecode inspection...  skip
Skipping kmidiff inspection...       skip
Skipping kmod inspection...          skip
Skipping license inspection...       skip
Skipping lostpayload inspection...   skip
Skipping lto inspection...           skip
Skipping manpage inspection...       skip
Skipping metadata inspection...      skip
Skipping modularity inspection...    skip
Skipping movedfiles inspection...    skip
Skipping ownership inspection...     skip
Skipping patches inspection...       skip
Skipping pathmigration inspection... skip
Skipping permissions inspection...   skip
Skipping politics inspection...      skip
Skipping removedfiles inspection...  skip
Skipping rpmdeps inspection...       skip
Skipping runpath inspection...       skip
Skipping shellsyntax inspection...   skip
[...]

PR #1331, opened 2024-01-29
[fix] Prevent a SIGSEGV in the unicode inspection

dcantrell:
Yeah, this can happen.  Add more checks for NULL.

Fixes: #1329


jimbair (approved):

msrb:
Thank you for the quick fix :)

PR #1332, opened 2024-01-30
Pin version of run-on-arch action for some fedora arches

rffontenelle:
The [version v2.7.1][v2.7.1] of uraimo/run-on-arch-action removed Dockerfiles for Fedora's s390x and ppc64le architecture (see https://github.com/uraimo/run-on-arch-action/commit/18a5cd7958b1f33f6a8f10b48fdbcbf02ccda003). Other Dockerfiles were removed, then readded (Fedora wasn't for some reason), but they will not be part of next v3.0.0.

So, either pin specific version for the action or remove these workflow files. This PR's commit proposes to pin the version.

See my test workflow runs: [fedora-ppc64le.yml][ppc64le] and [fedora-s390x.yml][s390x]

[v2.7.1]: https://github.com/uraimo/run-on-arch-action/commits/v2.7.1
[ppc64le]: https://github.com/rffontenelle/rpminspect/actions/runs/7704873344/job/20997913888
[s390x]: https://github.com/rffontenelle/rpminspect/actions/runs/7704872976

### DIFF
--- a/.github/workflows/fedora-ppc64le.yml
+++ b/.github/workflows/fedora-ppc64le.yml
@@ -39,10 +39,11 @@ jobs:
 
         # All of these steps run from within the main source
         # directory, so think of that as your $PWD
+        # NOTE: Pinned action version v2.7.0 because 2.7.1 removes fedora ppc64le
         steps:
             # This means clone the git repo
             - uses: actions/checkout@v4
-            - uses: uraimo/run-on-arch-action@v2
+            - uses: uraimo/run-on-arch-action@v2.7.0
               id: buildtest
               timeout-minutes: 500
               with:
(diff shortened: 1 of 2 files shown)


dcantrell:
This seems reasonable.  I wish s390x and ppc64le were available in a better capacity so that the entire test suite can run.  These CI jobs really just makes sure it builds, but that's still better than nothing.

dcantrell (changes requested):
The changes look fine, but can you modify the commit message so it follows what I have in CONTRIBUTING.md?  That is, give the short log message a prefix such as [ci] unless you do not want it to appear in the release notes.

Also, I do not mention this explicitly in the CONTRIBUTING.md file, but I prefer the long commit messages to be wrapped at somewhere between 72 and 80 columns.  I know we have large terminals now, but I find text like the git log and source code comments are actually easier to read when wrapped (think magazine or newspaper articles that use column formatting).

Thanks!

rffontenelle:
Commit message fixed. In future commits I'll pay more attention to it. Thanks for bringing it up.

dcantrell (approved):

PR #1333, opened 2024-02-01
Fix misspellings

rffontenelle:
I also found typos in CHANGES.md, but I see it was programmatically generated from git history. Should I fix them as well or the scripts would replace it?


dcantrell:
Feel free to fix the errors in CHANGES.md.  When I do a release, only the block for the new release is automatically generated.  The errors tend to come because I type way to fast and do not check my mistakes.  :)

PR #1343, opened 2024-03-15
Implement runtime remedy string overrides

dcantrell:
Requested by some users.  Allow the vendor data package to carry an optional config file that provides custom remedy strings.  rpminspect will still default to what it has now, but if you provide an override for a string it will use that one.

The remedy override file is explained in data/generic.yaml and the example file is in data/remedy/generic.toml


jimbair (approved):
Overall this looks good; the centralizing of the key values into one place and simplifying the code is a wonderful byproduct of the added remedfile toml support being added.

Assuming CI behaves itself with either a retry or a follow-up bugfix commit, LGTM.

dcantrell:
Yeah, I've gotta update the CI files so it sets up the environment correctly.  That's coming next.

PR #1347, opened 2024-03-19
Embed libtoml in source tree for systems that lack it

dcantrell:
This is easier than chasing it down on every non-Fedora platform and manually installing it.  The project has not changed upstream in about 8 years anyway.


jimbair (approved):
This seems fine to me considering the stability of the upstream release. +1

PR #1354, opened 2024-03-26
[ci] Remove Slackware Linux from the CI pool

dcantrell:
This job just stopped working and I can get it to work on either Slackware's latest stable release or the current development release. Patches welcome.


jimbair:
I would consider filing an issue to say it needs to be fixed, but I am fine to disable it for now.

jimbair (approved):

PR #1355, opened 2024-03-27
Exclude libpython3.so from the debuginfo tests

hroncok:
libpython3.so doesn't contain any compiled code
and it's there to point applications to the versioned libpython library where the compiled code is. Hence we cannot strip debug symbols from there.

See also: https://bugzilla.redhat.com/show_bug.cgi?id=2144553#c16

Cherry-picked from https://github.com/rpminspect/rpminspect-data-fedora/commit/fb53cacddcc8f72ee475b5e83b8d41851cbd7c85

This does not only affect Fedora.

### DIFF
--- a/data/generic.yaml
+++ b/data/generic.yaml
@@ -1079,6 +1079,7 @@ debuginfo:
         - /lib/modules/*
         - /usr/lib/debug/.dwz/*
         - /usr/lib/grub/*
+        - /usr/lib/debug/usr/lib*/libpython3.so*
 
     # The ELF section name(s) required in debuginfo packages.  The
     # default is shown here.  Some products may need to modify this


dcantrell:
I can add this for the example config files, but if you have not submitted changes to rpminspect-data-redhat or rpminspect-data-centos, you need to do that too.

hroncok:
I assumed this is the default config :/

Where can I add this to apply everywhere? Do I need to add it to fedora (already done), redhat, centos, each of them separately?

dcantrell:
Yeah, it has to be applied to each vendor data package separately.  There is no default config file set.  There are internal defaults for things, but this particular setting the default is an empty list.

The reason I don't have a default config and pushed everything to vendor data packages is to reinforce the idea that the config files make up distribution policy which rpminspect does not own.

hroncok:
I'll do that.

(This particular thing should not be solved by policy but by detecting the case in code, but I don't have enough resources to do that.)

dcantrell:
I would love to fix this in the code and it's one of those things on my long term to do list as well, but for now the config files can drive the right thing.

hroncok:
centos: https://gitlab.com/redhat/centos-stream/ci-cd/rpminspect-data-centos/-/merge_requests/42

hroncok:
redhat internal at osci/rpminspect-data-redhat MR 145

PR #1359, opened 2024-04-05
[fix] Drop libtoml-devel BR from the spec file

msrb:
This BR is no longer needed as the libtoml source code has been added to the repository in 6577ec61.

This should fix the FTBFS problem for EPEL buildroots in COPR.


dcantrell:
Oh yeah, oops.

dcantrell (approved):

Issue #1194, opened 2023-05-31
find_file_peers() has quadratic behavior, for compares of kernel packages it takes tens of minutes


dvlasenk:
I added this debug printout in lib/peers.c
```
        /* match up file peers between builds */
        if (peer->before_files && peer->after_files) {
fprintf(stderr, "find_file_peers >\n");
            find_file_peers(peer->before_files, peer->after_files);
fprintf(stderr, "find_file_peers <\n");
        }
```
and when I run `rpminspect-fedora -p rawhide -Dv -k -t VERIFY -w $PWD/TEMPDIR -a x86_64,noarch,src kernel-6.3.0-0.rc7.56.fc39 kernel-6.3.0-63.fc39` with pre-downloaded kernel-* directories, I see this (the lines with dots are the output of a performance monitoring tool):
```
find_file_peers >
00:29:52 U.......................................................................................    0    0
....
00:39:16 S.......................................................................................    0 8192
00:39:17 S.......................................................................................    0    0
00:39:18 S.......................................................................................    0    0
find_file_peers <
```
IOW: it spends ~10 minutes there. It would probably run into ~50 minutes if the compare wouldn't be limited to just x86_64.

I dug deeper and added this:
```
static Elf *get_elf_with_kind(const char *fullpath, int *out_fd, Elf_Kind kind)
{
....
    elf = elf_begin(fd, ELF_C_READ_MMAP_PRIVATE, NULL);
fprintf(stderr, "get_elf_with_kind('%s'): elf_kind(elf):%d kind:%d\n", fullpath, (int)elf_kind(elf), (int)kind);
```
I found out that through
```
bool is_elf(const char *path)
{
    return (is_elf_file(path) || is_elf_archive(path));
}
```
the get_elf_with_kind() gets called very, very, VERY many times. It translates to lstat()+open()+mmap()+close() per each call. This is why it takes 10 minutes to run all find_file_peers().

For example, (randomly picking a file), before/x86_64/usr/src/debug/kernel-6.3-rc7/linux-6.3.0-0.rc7.56.fc39.x86_64/arch/x86/mm/ident_map.c had been get_elf_with_kind()'ed  3284 times.

In total, the comparison run has 262076 extracted files, and there were 138342270 (138 million) get_elf_with_kind() calls.

I suggest: in extract_rpm(), in this code:
```
        /* Write the file to disk */
        if (archive_read_extract(archive, entry, archive_flags) != ARCHIVE_OK) {
[...]

PR #1361, opened 2024-04-07
Cache the result of "is it an ELF file?" checks

dvlasenk:
Add four flag fields to rpmfile_entry_t struct to save the information such as "is it an ELF archive?"
"is it ELF file?" and if yes, is it executable, or shared lib. (the fields are "signed char" to keep the struct compact, we have 100 thousands of those when running on kernel packages).

get_elf() attempts to open the file only if
"is it ELF file?" flag is "don't know".
It sets the flag to yes/no.
Future get_elf()'s on file which is known to not be an ELF immediately return NULL with no file operations attempted.

get_elf_archive() does the same with "is it an ELF archive?" flag.

This is the key speedup. It stops us from opening files and testing them for ELFness some 100+ million times for kernel packages.

is_elf_file() uses get_elf()'s cached result if it is present (therefore does not open the ELF not only if it's known to not be an ELF, but even if it *is* an ELF). Else calls get_elf() to find out.

is_elf_archive() uses get_elf_archive()'s cached result if it is present, else calls get_elf_archive().

is_elf_shared_library() uses get_elf() but additionally sets "is it ELF shared lib flag?" and does not try get_elf() on the second call, since it already has the cached answer. is_elf_executable() is similar, for executables.

### DIFF
--- a/lib/readelf.c
+++ b/lib/readelf.c
@@ -111,34 +111,46 @@ static Elf *get_elf_with_kind(const char *fullpath, int *out_fd, Elf_Kind kind)
 /* Return a read-only Elf object for the given path, or NULL on error or if the path is not an ELF object.
  * On success, the opened file descriptor is written to out_fd.
  */
-Elf * get_elf(const char *fullpath, int *out_fd)
+Elf * get_elf(rpmfile_entry_t *file, int *out_fd)
 {
-    return get_elf_with_kind(fullpath, out_fd, ELF_K_ELF);
+    /* -1 "no", 1 "yes", 0 "don't know yet" */
+    if (file->is_elf_file == -1) {
+        return NULL;
+    }
+
+    Elf *r = get_elf_with_kind(file->fullpath, out_fd, ELF_K_ELF);
+    file->is_elf_file = r ? 1 : -1;
+    return r;
 }
 
 /* Like above, but verifies that the file is an archive instead of an ELF file */
-Elf * get_elf_archive(const char *fullpath, int *out_fd)
+Elf * get_elf_archive(rpmfile_entry_t *file, int *out_fd)
 {
-    return get_elf_with_kind(fullpath, out_fd, ELF_K_AR);
+    /* -1 "no", 1 "yes", 0 "don't know yet" */
+    if (file->is_elf_archive == -1) {
+        return NULL;
+    }
+
+    Elf *r = get_elf_with_kind(file->fullpath, out_fd, ELF_K_AR);
+    file->is_elf_archive = r ? 1 : -1;
+    return r;
 }
 
 /*
  * Return true if a specified file is ELF, false otherwise.
  */
-bool is_elf(const char *path)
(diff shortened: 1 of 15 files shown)


lib/readelf.c
```diff
@@ -153,30 +163,46 @@ static bool _is_elf_type(const char *path, int type)
  * Return true if a specified file is an ELF shared library file, that
  * is, of type ET_DYN.
  */
-bool is_elf_shared_library(const char *path)
+bool is_elf_shared_library(rpmfile_entry_t *file)
 {
-    return _is_elf_type(path, ET_DYN);
+    /* -1 "no", 1 "yes", 0 "don't know yet" */
+    if (file->is_elf_shared_library != 0)
```
dcantrell, line 169: Please use { } around the block of this if statement even if it is only a single line.

lib/readelf.c
```diff
@@ -153,30 +163,46 @@ static bool _is_elf_type(const char *path, int type)
...
 /*
  * Return true if a specified file is an ELF executable file, that
  * is, of type ET_EXEC.
  */
-bool is_elf_executable(const char *path)
+bool is_elf_executable(rpmfile_entry_t *file)
 {
-    return _is_elf_type(path, ET_EXEC);
+    /* -1 "no", 1 "yes", 0 "don't know yet" */
+    if (file->is_elf_executable != 0)
```
dcantrell, line 184: Please use { } around the block of this if statement even if it is only a single line.

lib/readelf.c
```diff
@@ -153,30 +163,46 @@ static bool _is_elf_type(const char *path, int type)
...
 }
 
 /*
  * Return true for any ELF file.
  */
-bool is_elf_file(const char *path)
+bool is_elf_file(rpmfile_entry_t *file)
 {
+    /* -1 "no", 1 "yes", 0 "don't know yet" */
+    if (file->is_elf_file != 0)
```
dcantrell, line 198: Please use { } around the block of this if statement even if it is only a single line.

dcantrell (changes requested):
Thank you for this PR.  I like the approach.  Just one small request regarding coding style and that's to always use { } around if and for blocks.  I only saw three.  Thanks!

dcantrell:
Oh, also in the commit message can you put either:

Fixes: #1194 

or:

Resolved: #1194

On a line above the Signed-Off line?  See https://github.com/rpminspect/rpminspect/commit/cbc4d36c5d319fa88f8f2d67cb8e4cd987afca33 as an example.

lib/readelf.c
dvlasenk: done

lib/readelf.c
dvlasenk: done

lib/readelf.c
dvlasenk: done

dcantrell:
Thanks for the updates, I am running through some local tests and will likely have this merged today.

dcantrell (approved):

PR #1363, opened 2024-04-09
[fix] Build against the system libtoml in Fedora

msrb:
This should fix the following error in the build log:

```
Checking for function "toml_init" : NO 
Program pic_bits.sh found: YES (/builddir/build/BUILD/rpminspect-2.0-202404081240git3d3bff26/lib/pic_bits.sh)

lib/meson.build:176:14: ERROR: Unknown variable "toml".
```

### DIFF
--- a/rpminspect.spec.in
+++ b/rpminspect.spec.in
@@ -50,6 +50,9 @@ BuildRequires:  libmandoc-devel >= 1.14.5
 BuildRequires:  gnupg2
 BuildRequires:  libicu-devel
 BuildRequires:  libcdson-devel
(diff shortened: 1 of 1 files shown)


msrb:
This fixes the build for Fedora:
https://copr.fedorainfracloud.org/coprs/msrb/rpminspect/build/7281038/

Epel is still broken:
https://download.copr.fedorainfracloud.org/results/msrb/rpminspect/epel-9-x86_64/07281038-rpminspect/builder-live.log.gz

dcantrell (approved):

PR #1364, opened 2024-04-10
[bld] Generate .tar.gz source archives rather than .tar.xz

dcantrell:


jimbair (approved):
What's old is new again

PR #1365, opened 2024-04-10
[bld] Build libtoml sources without -Werror

msrb:
COPR builds for EPEL fail because the sources are compiled with -Werror. The problem is that the bundled libtoml sources don't compile when warnings are treated as errors.

The right thing to do would be to work with the libtoml upstream to fix the problems and make sure that they use -Werror as well. But this can take some time to achieve.

In the meantime, we compile the libtoml sources without -Werror.

### DIFF
--- a/libtoml/meson.build
+++ b/libtoml/meson.build
@@ -10,4 +10,5 @@ libtoml = static_library(
     libtoml_sources,
     include_directories : inc,
     install : false,
+    override_options: ['werror=false']
 )


dcantrell (changes requested):
Change the libtoml asprintf() usage to xasprintf() as provided in the rpminspect headers.  Keep -Werror in place.

msrb:
@dcantrell I was thinking about just patching the bundled code, but do you expect any upstream rebases there?

msrb:
The last upstream commit is from 8 years ago... so I guess there won't be any "rebase to the latest upstream release" commits here :)

In that case, patching the code here is indeed the best thing to do -- I am on it ;)

msrb:
Superseded by https://github.com/rpminspect/rpminspect/pull/1366.

PR #1366, opened 2024-04-11
[fix] Replace asprintf() with xasprintf() in libtoml source code

msrb:
Compiler complains about the return value of `asprintf()` being ignored, and -Werror flag causes the build to fail because of that.

`xasprintf()` makes this warning to go away.

However, `xasprintf()` was previously defined in `include/rpminspect.h`, and there were conflicts between this header file and `ccan/list/list.h`. Therefore I moved the `xasprintf()` definition to a separate header file -- `include/helpers.h`. See the commit message for more details.

### DIFF
--- a/include/helpers.h
+++ b/include/helpers.h
@@ -1,0 +1,38 @@
+/*
+ * Copyright The rpminspect Project Authors
+ * SPDX-License-Identifier: LGPL-3.0-or-later
+ */
+
+/*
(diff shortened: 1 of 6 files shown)


msrb:
Test build in COPR: https://copr.fedorainfracloud.org/coprs/msrb/rpminspect/build/7296739/

PR #1368, opened 2024-04-11
[lib] annocheck: Map el10 product release to the corresponding profile

msrb:
The latest annobin release already supports el10.


dcantrell (approved):
So it is:
https://sourceware.org/cgit/annobin/commit/annocheck/hardened.c?id=1ad83a58e1a1d9376bcd5337dd1e827e2914132b

PR #1371, opened 2024-04-15
[ci] Fix up failing CentOS 7 and Oracle Linux CI jobs

dcantrell:
Also do a bit of shell script cleanup work.  For CentOS 7, had to adjust the way pip modules were installed and then python-rpm was rebuilt.  For Oracle, we're using 2 different versions of Python 3, but that's ok.


jimbair:
While CentOS 7's EOL approaches fast, it's not dead yet! I approve.

jimbair (approved):

PR #1384, opened 2024-05-06
[tst] LooseVersion from distutils.version is deprecated; replace it

dcantrell:
Use packaging.version's parse() and Version() instead.


jimbair (approved):
It's always good to move away from deprecated methods

PR #1390, opened 2024-05-13
[tst] Handle older Python 3.x releases without 'packaging' module

dcantrell:
I was using LooseVersion from distutils, but that's deprecated in more recent releases.  Since the CI jobs run on a variety, still handle cases where we should still use LooseVersion.

### DIFF
--- a/test/test_disttag.py
+++ b/test/test_disttag.py
@@ -9,19 +9,25 @@ import shutil
 import subprocess
 import tempfile
 import unittest
-from packaging.version import Version, parse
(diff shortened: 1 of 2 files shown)


jimbair:
I don't see any obvious way to import the second module using `import foo as bar` the first to avoid duplicating logic, so this seems fine. A comment would be helpful to explain it since it's a bit funny, even if it's mildly self evident it's a platform problem, but it's nice to explain that kind of thing I think.

jimbair (approved):

Issue #1387, opened 2024-05-13
Bogus "Mixed SPDX and legacy license identifiers" report: Python-2.0.1 AND CC0-1.0 AND MIT


hroncok:
**Describe the bug**

rpminspect reports an SPDX expression as "mixed SPDX and legacy license identifiers". I belive this is a valid SPDX expression.

    Python-2.0.1 AND CC0-1.0 AND MIT

**To Reproduce**

```json
{
  "diagnostics": [
    {
      "result": "DIAGNOSTICS",
      "message": "Version information for libraries and programs used by rpminspect as well as storage requirements.  This result is for informational and diagnostic purposes only.",
      "details": "rpminspect version 2.0\nzlib version 1.2.13\nlibmagic version 544\nlibclamav version 1.0.6\nlibrpm version 4.18.2\nlibxml version 21004\njson-c version 0.17\nlibcurl version 8.0.1 (OpenSSL\/3.0.9 zlib\/1.2.13 brotli\/1.0.9 libidn2\/2.3.7 libpsl\/0.21.2 (+libidn2\/2.3.4) libssh\/0.10.6\/openssl\/zlib nghttp2\/1.52.0 OpenLDAP\/2.6.6)\nlibarchive version 3.6.1 (zlib\/1.2.13 liblzma\/5.2.9 bz2lib\/1.0.8 liblz4\/1.9.4 libzstd\/1.5.2)\nlibyaml version 0.2.5\nOpenSSL version 3.0.9 30 May 2023\nxmlrpc-c version 1.51.8\nmsgunfmt (GNU gettext-tools) 0.21.1\nannocheck version 12.40.\nabidiff version 2.4.0\nkmidiff version 2.4.0\nudevadm version 253\n\nSpace required to download artifacts: 0 bytes (0 B)\nSpace required to unpack artifacts: 681719603 bytes (650 MiB)\n"
    },
    {
      "result": "DIAGNOSTICS",
      "message": "Command line arguments used to invoke rpminspect.",
      "details": "\/usr\/bin\/rpminspect -c \/usr\/share\/rpminspect\/fedora.yaml -Dv -k -t BAD -w \/var\/tmp\/rpminspect -a x86_64,noarch,src -o \/root\/zuul-output\/logs\/result.json -F json \/var\/tmp\/rpminspect\/python3.13-3.13.0~b1-1.fc41 \/var\/tmp\/rpminspect\/python3.13-3.13.0~b1-2.fc41"
    },
    {
[...]

PR #1394, opened 2024-05-16
[fix] Correctly handle dual SPDX/legacy license identifiers

dcantrell:
The license inspection was incorrectly handling license identifiers that appear as both legacy identifiers in the license database as well as SPDX expressions.  The core problem is that under the legacy system, there was a lot of identifier reuse.  This patch refactors how the license inspection notes the dual licenses so that the inspection passes on packages that have migrated to SPDX.

Fixes: #1387

### DIFF
--- a/lib/init.c
+++ b/lib/init.c
@@ -599,28 +599,17 @@ static inline void _remedy_walker(struct toml_node *node, void *data)
         r = toml_value_as_string(node);
 
         if (r) {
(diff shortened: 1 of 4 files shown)


jimbair (approved):
As per our discussion, all looks good here (mostly just me asking via DM about C syntax). Hopefully this fixes up the issue!

PR #1404, opened 2024-05-31
[lib] reduce struct size of rpmfile_entry_t

dvlasenk:
On x86_64, struct stat st is 128 bytes, but we use ony three fields in it. Replacing it with
    off_t st_size;
    mode_t st_mode;
    unsigned st_nlink;
in rpmfile_entry_t reduces the latter from 240 bytes to 112 bytes.

In test run a-la
rpminspect -c /usr/share/rpminspect/fedora.yaml -p rawhide
    -Dv -k -t VERIFY -a x86_64,noarch,src -T virus kernel-6.8.10-200.fc39
after rpm unpacking memory consupmtion of the process fell from ~130 to ~100 megabytes.

### DIFF
--- a/include/types.h
+++ b/include/types.h
@@ -108,7 +108,10 @@ typedef struct _rpmfile_entry_t {
     Header rpm_header;
     char *fullpath;
     char *localpath;
-    struct stat st;
+    /* struct stat st; - 128 bytes on x86, store only fields we need: */
+    off_t st_size;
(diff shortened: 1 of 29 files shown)


dcantrell (approved):

dcantrell:
The failing tests are due to unrelated changes (rpmfluff is outdated on rawhide).

PR #1436, opened 2024-10-09
[lib] Disable file and scan size limits in clamav

dcantrell:
By default libclamav enforces file size and scan size limits.  I assume this is to cover most file types by default and not appear to hang or stall.  We want to scan all files and some files that appear in RPMs may be beyond that limit.  So disable those limits when we run the virus inspection.


jimbair (approved):
Go big or go home

PR #1450, opened 2024-11-22
Two fixes for librpminspect

dcantrell:
1) Make sure unapproved licenses are reported as such

Ran in to a problem where unapproved licenses were showing up as approved.  Not good.

2) Refactor how remedy strings are handled in librpminspect

If a config file provided remedy string overrides, these were getting lost and we hit a NULL pointer dereference in the reporting.  Refactored how remedy strings are handled with the defaults and then the user overrides.  We now get the expected remedy string in the results.

### DIFF
--- a/include/remedy.h
+++ b/include/remedy.h
@@ -8,6 +8,7 @@ extern "C"
 {
 #endif
 
+#include <stdbool.h>
 #include "i18n.h"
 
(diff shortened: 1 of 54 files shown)


jimbair (approved):
Bug fixes always welcomed, assuming CI passes of course.

PR #1506, opened 2025-08-20
[lib] Refactor how %doc and %license lines are processed in filesmatch

dcantrell:
What I originally wrote was a bit too long and difficult to follow so I shortened it and made a helper function.

### DIFF
--- a/lib/inspect_filesmatch.c
+++ b/lib/inspect_filesmatch.c
@@ -133,6 +133,47 @@ static char *trim_files_modifiers(char *s)
 }
 
 /*
+ * Process %doc and %license lines in %files blocks.
+ */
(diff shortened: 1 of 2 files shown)


lib/inspect_filesmatch.c
```diff
@@ -132,6 +132,47 @@ static char *trim_files_modifiers(char *s)
...
+    char *p = NULL;
+    string_entry_t *token = NULL;
+
+    assert(name != NULL);
+    assert(tokens != NULL);
+    assert(macro != NULL);
+    assert(path != NULL);
+
+    /* directory prefix */
+    xasprintf(&prefix, "%%{%s}/%s", macro, name);
```
jimbair, line 150: Should the / here be PATH_SEP? Or is the / fine?

PR #1509, opened 2025-08-20
[lib] Hook up the actual filesmatch inspection

dcantrell:
The data has been collected from the spec file and the RPMs are unpacked, so check what's in the package against what the spec file rules are.  Report any problems.


jimbair (approved):

jimbair (approved):
Looks good over here; I look forward to grinding through our build list.

PR #1589, opened 2026-05-26
[fix] Do not scan for a possible 'exec' line if shell is NULL

dcantrell:
In the shellsyntax inspection, do not continue reading the script for a possible exec line if we did not find a #! shell line.  This fixes up some false detection of text files that say "exec sh" but are otherwise not shell scripts.  The idea of the 'exec' line detection is to ensure the correct shell is used to check the syntax of the script.


jimbair (approved):

sgallagher:
I disagree with this, as after this patch it will no longer inspect shell scripts that aren't executed directly, but are instead imported via `source` or `.` directives. Could we maybe run `bash -n <file>` first to see if it parses successfully without actually executing anything? That way, we can scan anything that is valid shell script, rather than just those things that have the hashbang.

dcantrell:
There's not really a good way to distinguish between "bad shell script" and "not a shell script".  I looked at using ShellCheck which gives a little more options for checking shell scripts, but is limited to sh.  That may be ok for the purposes of advancing through the shellsyntax inspection.

What was happening is that those Tcl scripts in dejagnu were getting marked as "sh" scripts since it found "exec sh" in the script.  I look at the MIME type from libmagic, but get "text/plain" so I can really only match "text/" for scripts.  You only get "text/x-shellscript" if the script has a #! line at the top.

Before this patch it was running "sh -n" on the remote.exp file and reporting that as a bad shell script.  Because there's no way that I can distinguish not a shell script from shell script.

sgallagher:
> Because there's no way that I can distinguish not a shell script from shell script.

Ask an LLM to do it for you? (I'm only half-kidding: a small model probably could come up with a reasonable heuristic determination.)

dcantrell:
tools

